PR #3, opened 2015-06-11
Custom display control

umupfumu:
Adds a custom display control that takes in a template URL and fills it with data from observations.  Also adds a birth certificate for the possible config that uses the custom display control.


umupfumu:
Looks like this breaks the orders dispay control?

umupfumu:
Seems like a strange problem w/ orders date/time.  not sure why this commit would have any affect on that.

umupfumu:
@shruthidipali @maheshongithub -- i heard from anant there was some progress on this? any way i can help?

PR #14, opened 2016-06-03
fix error in registration when editing patient with free-text value for address field mapped to hierarchy

mogoodrich:


bharatak:
@mogoodrich Can you please squash the commits? Also, can you please add a testcase for this scenario in topDownAddressFields.spec.js.  Not sure if package.sh change is required

mogoodrich:
@bharatak sure, I can look into squashing the commits.  Not sure exactly how the other commits got picked up as included, but the net diff is only in the topDownAddressField file.

PR #17, opened 2016-06-06
#1755: Ability to configure forward-url on click of Save on Registration visit page

mogoodrich:
See:

https://bahmni.mingle.thoughtworks.com/projects/bahmni_emr/cards/1755

and

 https://talk.openmrs.org/t/save-visit-details-and-forward-url/6340/3

I've added a new  registration app config parameter "afterVisitSaveTransitionToState" where you can specify which state to transition to after saving.  If the parameter is not specified, the current behavior (stay on same page) is maintained.

### DIFF
--- a/ui/app/registration/controllers/visitController.js
+++ b/ui/app/registration/controllers/visitController.js
@@ -163,17 +163,21 @@ angular.module('bahmni.registration')
             };
             //End :: Registration Page validation
 
-            var reload = function () {
-                $state.transitionTo($state.current, $state.params, {
+            var afterSave = function () {
+
+                var nextState = appService.getAppDescriptor().getConfigValue('afterVisitSaveTransitionToState');
+
+                $state.transitionTo(nextState ? nextState : $state.current, $state.params, {
                     reload: true,
                     inherit: false,
                     notify: true
                 });
+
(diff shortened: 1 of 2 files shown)


sravanthi17:
The approach you have taken is to always accept "state" as a value of "afterVisitSaveTransitionToState"

In this case,

1) We expect the implementer to have good knowledge about angular states. We don't really want to expose our states.

2) The states one can configure are always limited with in the registration app.

For example, the implementer want to navigate to patient dashboard which is not possible in the approach followed.

Otherwise i would suggest asking the implementer to provide **Transition url** or **Forward url** (url to navigate to after save). So we don't need to depend on the states.
Accepting the forward url from the implementer is already followed in Bahmni.

One example is https://github.com/Bahmni/openmrs-module-bahmniapps/blob/master/ui/app/registration/directives/patientAction.js#L80

Also the configuration should be from "**extension.json**" to be consistent across.

mogoodrich:
Thanks @sravanthi17 

Using a Url instead of a state makes sense to me.  I will close this out and issue a new pull request.

Not sure I agree 100% on app.json vs extension.json... will comment in the new pull request.

PR #21, opened 2016-06-08
#1755: Ability to configure forward-url on click of Save on Registration visit page

mogoodrich:
This is an updated pull request based on feedback from previous pull request.  I have changed the "afterVisitSaveTransitionToSave" configuration parameter to "afterVisitSaveForwardUrl", which I agree is more flexible and closer in line with what we are doing elsewhere.

However, I kept this in the registration app.json, not extension,json, because this is not an extension ponit we are configuring, and it seemed needlessly complex just to add an extension point for this.  I follow the pattern for "searchByIdForwardUrl", which is in app.json, not extension.  I also noted that other "forward" urls are configured in app.json in other locations, for instance the ADT app.

Let me know if you have further questions/comments, thanks!
Mark

### DIFF
--- a/ui/app/registration/controllers/visitController.js
+++ b/ui/app/registration/controllers/visitController.js
@@ -166,7 +166,7 @@ angular.module('bahmni.registration')
             var afterSave = function () {
                 var forwardUrl = appService.getAppDescriptor().getConfigValue("afterVisitSaveForwardUrl");
                 if (forwardUrl != null) {
-                    $location.url(appService.getAppDescriptor().formatUrl(forwardUrl, {'patientUuid': patientUuid}));
+                    $window.location.href = appService.getAppDescriptor().formatUrl(forwardUrl, {'patientUuid': patientUuid});
                 }
                 else {
                     $state.transitionTo($state.current, $state.params, {
--- a/ui/app/registration/controllers/visitController.js
+++ b/ui/app/registration/controllers/visitController.js
@@ -163,17 +163,23 @@ angular.module('bahmni.registration')
             };
             //End :: Registration Page validation
 
-            var reload = function () {
-                $state.transitionTo($state.current, $state.params, {
-                    reload: true,
-                    inherit: false,
-                    notify: true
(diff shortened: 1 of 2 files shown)


ui/app/registration/controllers/visitController.js
```diff
@@ -163,17 +163,23 @@ angular.module('bahmni.registration')
...
-            var reload = function () {
-                $state.transitionTo($state.current, $state.params, {
-                    reload: true,
-                    inherit: false,
-                    notify: true
-                });
+            var afterSave = function () {
+                var forwardUrl = appService.getAppDescriptor().getConfigValue("afterVisitSaveForwardUrl");
+                if (forwardUrl != null) {
+                    $location.url(appService.getAppDescriptor().formatUrl(forwardUrl, {'patientUuid': patientUuid}));
```
sravanthi17: $location.url has some problem. when you configure url within different app, it navigates to root url within the app. you can reproduce it by configuring patient dashboard url.

mogoodrich: Got it, thanks.  Just to clarify, is there an existing pattern/method I should be using? I see that in the example you linked to in the previous pull request was using $window.location.href. Is that what I should be using here? Or is a new issue specific to this use case that I will need to work around?  Thanks!

sravanthi17: "$window.location.href"  works fine in this case.

mogoodrich: Great, thanks, I have updated this pull request with a further commit that changes to $window.location.href

mogoodrich:
Whoops, duh... forgot to update the unit test after the change to $window.location.href.  Hold on this one, I will update and issue a new request.

mogoodrich:
Okay, see updated pull request here with change to $window.location.href, fixed test, and everything squashed into a single commit.

https://github.com/Bahmni/openmrs-module-bahmniapps/pull/22/files

I will close this pull request.

PR #23, opened 2016-06-19
fix package.sh test to see if Xvfb is running

mogoodrich:
The test to see if Xvfb is running does not appear to be working correctly.  See my inline comments.

### DIFF
--- a/ui/scripts/package.sh
+++ b/ui/scripts/package.sh
@@ -13,12 +13,13 @@ rm -rf $ROOT_DIR/target/${ZIP_FILE_NAME}*.zip
 npm install
 bower install
 
-pids=$(pgrep Xvfb)
-if [ -n "$pids" ]; then
+if [ -z $(pgrep Xvfb) ]; then
     export DISPLAY=:99
     Xvfb :99 &
+    XVFB_PID=$!
+    echo "Starting Xvfb process $XVFB_PID"
 else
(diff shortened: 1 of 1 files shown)


ui/scripts/package.sh
```diff
@@ -13,8 +13,7 @@ rm -rf $ROOT_DIR/target/${ZIP_FILE_NAME}*.zip
 npm install
 bower install
 
-pids=$(pgrep Xvfb)
```
mogoodrich: If Xvfb is not running, the exit code of the pgrep will be 1, and therefore the exit code of the variable assignment is also 1. Since "set -e" is set earlier in the commit, the script will always exit at this point.

ui/scripts/package.sh
```diff
@@ -13,8 +13,7 @@ rm -rf $ROOT_DIR/target/${ZIP_FILE_NAME}*.zip
...
-if [ -n "$pids" ]; then
+if [ -z $(pgrep Xvfb) ]; then
```
mogoodrich: Putting the pgrep directly within the test prevents a non-zero exit code from prematurely terminating the script.

Also note that I believe we should be testing with -z here, not -n, since we want to return true if the value DOES have zero size.

mogoodrich:
Have updated, will issue new pull request, closing this one out.

PR #24, opened 2016-06-19
fix package.sh to properly start and stop Xvfb if necessary

mogoodrich:
package.sh didn't appear to be properly starting and stopping Xvfb. See comments inline.


ui/scripts/package.sh
```diff
@@ -13,12 +13,13 @@ rm -rf $ROOT_DIR/target/${ZIP_FILE_NAME}*.zip
 npm install
 bower install
 
-pids=$(pgrep Xvfb)
```
mogoodrich: If Xvfb is not running, the exit code of the pgrep will be 1, and therefore the exit code of the variable assignment is also 1. Since "set -e" is set earlier in the commit, the script would always exit as this point.

ui/scripts/package.sh
```diff
@@ -13,12 +13,13 @@ rm -rf $ROOT_DIR/target/${ZIP_FILE_NAME}*.zip
...
-if [ -n "$pids" ]; then
+if [ -z $(pgrep Xvfb) ]; then
```
mogoodrich: Putting the pgrep directly within the test prevents a non-zero exit code from prematurely terminating the script.

Also note that I believe we should be testing with -z here, not -n, since we want to return true if the value DOES have zero size.

ui/scripts/package.sh
```diff
@@ -31,3 +32,9 @@ cd $ROOT_DIR/dist && zip -r ../target/${ZIP_FILE_NAME}_chrome.zip *
 cd ..
 grunt android
 cd $ROOT_DIR/dist && zip -r ../target/${ZIP_FILE_NAME}_android.zip *
+
+if [ -n $XVFB_PID ]; then
+    echo "Killing Xvfb process $XVFB_PID"
+    kill $XVFB_PID
```
mogoodrich: We also need to make sure we kill the Xvfb process if we started it--otherwise this script does not exit.

sravanthi17:
@mogoodrich : Thanks for your contribution. We merged this commit and will be available from 0.83v

mogoodrich:
Thanks!

PR #26, opened 2016-07-23
Fixing tests that fail when run outside of IST time zone

mdjnewman:


sravanthi17:
Merged this in 0.84v of Bahmni. Thanks for your contribution :)

PR #27, opened 2016-08-09
bindu | Added Portuguese Language locale files

binduak:
screen shot 2016-08-09 at 1 26 12 pm


padmavati:
Hi Bindu,
Thanks for contributing to Bahmni. 
1)Modules in the home page are not translated.Please refer to the attached screenshot.
2)Location translation version of portuguese is not correct.
3)Tabs under clinical  module also not translated. 
4)Tabs under Orders module are not translated.
screen shot 2016-08-10 at 11 24 11 am

gsluthra:
Whats the update on this pull request? Do the home module elements need translation? Can we have some logic in code which picks up "English" versions, if it doesn't find the Portuguese version?

padmavati:
We don't have support for if Portuguese version not available fallback to English. I will test it and merge if it looks fine.

gsluthra:
I got to know today that these strings are in the "default-config", and hence won't be translated in Transifex (since these are not product strings). The idea will be that people who are deploying Bahmni, can configure these strings themselves. So, its ok, and we should go ahead and merge the pull requests if everything else looks fine.

padmavati:
Hi,

The above pull request is having the following issues:
1)In the clinical dashboard since some of the strings are not having the translation for portuguese language, UI is breaking.However all of them are clickable.

screen shot 2016-08-19 at 5 51 00 pm

2)In patient dashboard after entering the consultation, all the tabs are overlapping(UI is breaking).In the following screenshot **History and Examinations**  and **Vitals** are not clickable. Same happens for other tabs as well. All the content below tabs is not clickable.

screen shot 2016-08-19 at 5 51 29 pm

gsluthra:
I would suggest to have a screenshare with Nilo, to show him how Bahmni in Portuguese looks. That way we will also get timely feedback, in case something isn't right and needs to be fixed. Also, I was wondering whether its possible in the same call to also update "default-config" strings to add Portuguese strings (keeping English as-is), so that default config works fine in English and also shows valid strings in Portuguese? Atleast for the dashboard, and tabs.

gsluthra:
Some sample screenshots attached (the strings in English are all data entered by user, so that is ok)

image

image

image

mahsudbek10:
hello everyone. 
how to translate attributes? for example: debt, class?
how to traslate new attributes where added me?

RafaNetow:
Hi everyone.
I would like to do a pull request and add Spanish to the project. And I would like to know if you could give me some advice as to what files and settings I should take into account for this change @gsluthra

sumanmaity112:
Hi @RafaNetow, you can check this [link](https://github.com/Bahmni/openmrs-module-bahmniapps/tree/master/ui/app/i18n). 
And can you please move this conversation to [OpenMRS talk](https://talk.openmrs.org/)?

PR #30, opened 2016-09-07
add grunt clean:coverage task

brookingcharlie:
If we don’t clean the coverage directory, we can get errors like this:

    Running "coverage" task
    Fatal error: Cannot read property 'length' of undefined

This is due to coverage reports being out of sync, as described in this
issue against the grunt-istanbul-coverage project:
https://github.com/daniellmb/grunt-istanbul-coverage/issues/10

### DIFF
--- a/ui/Gruntfile.js
+++ b/ui/Gruntfile.js
@@ -74,6 +74,9 @@ module.exports = function (grunt) {
                     }
                 ]
             },
+            coverage: [
+                'coverage'
+            ],
             debug: [
                 '<%= yeoman.app %>/styles/*.css'
             ]


bharatak:
@brookingcharlie Adding the coverage section is Gruntfile (as per your commit) is not removing the error.  As you mentioned, deleting the folder is temporarily solving the problem.

brookingcharlie:
@bharatak maybe I messed something up, but including this clean:coverage task should mean that doing a `grunt clean` removes the coverage/ folder. This is intended to help folks who get errors trying to run the tests and don't know that they need to delete coverage/ - doing a clean should ideally take care of it.

PR #33, opened 2016-10-20
move `bower install` into an npm `postinstall` script to simplify environment setup

epenkman:
this casues bower install to be triggered by `npm install`. 

This saves us having to run both `npm install` and `bower install` when package changes are made. 

`./node_modules/.bin/bower` is used so that the developer doesn't need`bower` installed globally.

I left in `grunt bower-install` because I don't know if it's being called externally to this project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ install:
   - gem install compass
   - cd ui && npm install
   - cat npm-debug.log || true
-  - bower install
 
 script: grunt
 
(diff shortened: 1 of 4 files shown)


vinkesh:
I was facing issue while running "npm install" with this change, because I have to run npm install on my machine with "sudo".

GautamSidtharth:
Post install script uses local instance of bower, but it is not installed via dev dependencies.
Hence, the below line will fail.
`"postinstall": "./node_modules/.bin/bower install"`
Would not it?

mks-d:
@endeepak I will close this for now. Please raise a ticket on https://bahmni.atlassian.net/ if this is still relevant.

PR #34, opened 2016-10-22
Travis build config changes

mdjnewman:
A few changes:

* Works around npm/npm#14042 using
https://github.com/travis-ci/travis-ci/issues/4653#issuecomment-194051953
* Print out the npm debug log for when npm fails
* Cache the npm and bower directories to speed up the build
* Explicitly disable email notifications
* Use 'dots' reporter rather than 'progress' for Karma to keep console output smaller

PR #36, opened 2016-10-23
Change ESLint config to match current codebase

mdjnewman:
* alert, prompt, and confirm are used in Bahmni (http://eslint.org/docs/rules/no-alert)
* no-empty-label is removed in ESLint 2
* wrap-iife (http://eslint.org/docs/rules/wrap-iife) was using the
  default 'outside' option, which isn't the standard being followed in
  Bahmni


mdjnewman:
See #34 for info about the failing Travis check.

Issue #37, opened 2016-10-23
ESLint vs JSHint


mdjnewman:
Hi!

Currently there is an .eslintrc file & a .jshintrc file checked in.

ESLint is used by Code Climate, whereas JSHint is used by the Grunt build.

There are also 2 errors & 571 warnings in the current .jshintrc output, so I'm wondering how much value it's providing (even errors don't fail the build).

I think it would be good to:
1. Standardise on one of these  (I'd prefer ESLint)
2. Chose what linting rules we actually care about, and make them fail the build
3. Remove all the other rules

Rules that just case warnings and don't break the build will be ignored most of the time, and just produce noise.

I'm happy to fix all the lint errors I can and change Grunt tasks if necessary, but I need some direction on which linter to use.  If we need ESLint to make use of Code Climate & want to keep the build using JSHint we should at least document why there are two, and configure them with similar rules, but I really think having just one is a better option.

Thoughts?

bharatak:
@mdjnewman Thanks a lot.  We would prefer to go by eslint.  Our other projects like [form-controls](https://github.com/Bahmni/form-controls) and [implementer interface](https://github.com/Bahmni/implementer-interface) use eslint.  They are ReactJS + ES6 projects and eslint is the standard.  We can standardize it across bahmni.  Its a good idea to follow [johnpapa's style guide](https://github.com/johnpapa/angular-styleguide/tree/master/a1).  Looks like there is an [npm package](https://www.npmjs.com/package/eslint-plugin-angular) implementing it.

I am thinking of the following tasks
- Change the Grunt tasks to use ESLint
- Fail the build on lint errors
- Have the styles that are compatible today and fix some very obvious errors.  If you have a big list to fix, I prefer multiple PRs and doing it slowly.  Just trying to be cautious.
- You can comment out (but still keeping it in .eslintrc) the rules that bahmni is not compatible with currently, but are good to have.  We can work on it as a tech debt to slowly incorporate one rule at-a-time.

Appreciate your efforts.

mdjnewman:
Thanks @bharatak, makes sense to standardise these things across projects. form-controls and implementer-interface use [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb) which we can also use as a base.

eslint-plugin-angular sounds good, and it's [supported by Code Climate](https://docs.codeclimate.com/docs/eslint). We can add rules from this plugin to the airbnb base.

[...]

PR #38, opened 2016-10-30
JSHint --> ESLint

mdjnewman:
This is part of Bahmni/openmrs-module-bahmniapps#37

Note that the huge diff in commit 276a779 was produced by the `eslint --fix` command, which automatically applies fixes for rules which don't change the AST (e.g. spaces/indentation).

6c13034 sets the error level on the failing rules to 'warn', so that the build isn't broken initially. As per the comments on the issue above, these can be removed from our .eslintrc when cleaned up to use the default rules from eslint-config-semistandard & eslint-config-angular.

cc @bharatak


mdjnewman:
Okay, I think I'm done on this for the weekend :) The last commit allows editors to warn about violated rules as per .eslintrc, but keeps the build output clean and makes it easier to see error level violations.

PR #40, opened 2016-11-11
Fixes for ESLint angular/di rule

mdjnewman:
Fixing violations of ESLint [angular/di](https://github.com/Gillespie59/eslint-plugin-angular/blob/master/docs/di.md) rule and enabling the rule at `error`level.

The rule has three alternatives:

* Function syntax
* Array syntax
* `$inject` syntax

Array syntax was chosen as it's in use across most of Bahmni, and is the preferred annotation option according to the [Angular docs](https://docs.angularjs.org/guide/di#dependency-annotation).

Most of the commits are fixing mismatches between parameter names and the elements in the array.

Some services were renamed to align with the [Factory and Service Names advice](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#style-y125) in  johnpapa's style guide.

The commit with message 'Convert all function syntax DI into array syntax' (b0dcac0 unless I rebase again) used a script to refactor all exiting annotations using function syntax to array syntax.

This is part of #37.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -37,6 +37,7 @@ globals:
 
 rules:
   angular/controller-as: off
+  angular/di: [error, array]
   indent: [error, 4]
   quotes: off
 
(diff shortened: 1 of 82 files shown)


mdjnewman:
Hi all, I've just rebased onto master.

shashikanthgadgay:
Hi @mdjnewman, I will merge this PR in couple of days.

mdjnewman:
Thanks for the update @shashikanthgadgay :)

PR #44, opened 2017-03-10
Conditions list Implementation. Show/Hide Patient middle name.

sidtharthanan:
Conditions list Implementation.
Show/Hide Patient middle name.


sidtharthanan:
We will raise two separate PRs.

PR #61, opened 2017-06-15
Any field in Audit Log filter filters when ENTER key is pressed

nzcoffeem:
https://trello.com/c/mYscHfzS/89-on-pressing-the-enter-key-in-any-filter-field-the-audit-log-filter-button-should-get-pressed


deepaucksharma:
Thank you for raising PR. Here is the mingle card to track the request - https://bahmni.mingle.thoughtworks.com/projects/bahmni_emr/cards/3627

PR #67, opened 2017-06-23
Introduced a toggle to hide conditions list.

mddubey:
The condition-list won't be shown on diagnosis tab if `bahmni_config/openmrs/apps/clinical/app.json` file has a property `config.hideConditions`  set as `true`.
The condition-list will be shown if this property is not present or set to `false`.


hanisha93:
When hideConditions is set to true, **add as condition** button is still showing up in current diagnosis. But the expected behaviour is to hide the the button along with conditions block.
screen shot 2017-07-07 at 11 19 28 am

mddubey:
Hey @hanisha93, Thanks. I will check that.

PR #76, opened 2017-09-19
Filter obs with obs select list

maharjunm:


ui/app/common/domain/services/observationsService.js
```diff
@@ -20,6 +20,9 @@ angular.module('bahmni.common.domain')
                 params.scope = scope;
                 params.patientProgramUuid = patientProgramUuid;
             }
+            if (obsSelectList) {
```
sumanmaity112: Why do we need this if condition?

sumanmaity112:
Hi @maharjunm @shireeshaBongarala,
As we don't need these changes, I am closing this PR.

PR #105, opened 2017-11-10
Last Visit Date on Registration > Edit Patient Details page

charl3sj:
Please note that this is a quick fix. A separate patient service was defined to pull out visit dates for a patient to populate the Last Visit Date field. Ideally, this data should come from the backend and be accessible from the $scope.patient object.


charl3sj:
This is a bad fix. After further discussion at Samanvay, we've decided to withdraw this pull request and submit a better one when time permits.

PR #108, opened 2017-11-22
BAH-380 REST API call to delete image or video file should be ca…

pramidatumma:
…lled only if the obs type is video or image and the obs have a value


ShruthiPitta:
@pramidatumma  Is this already merged? Can you please close this PR if so?

ShruthiPitta:
Confirmed with @pramidatumma This is redundant now. Hence closing it.

PR #112, opened 2017-12-08
BAH-398 | [Fix] Causes error If value for a key is undefined in Cookie

sai789:
JIRA id:
https://bahmni.atlassian.net/browse/BAH-398

### DIFF
--- a/ui/app/common/util/init.js
+++ b/ui/app/common/util/init.js
@@ -25,6 +25,7 @@ angular.module('bahmni.common.util', [])
                     if (jsonCookie) {
                         return angular.fromJson(decodeURIComponent(jsonCookie));
                     }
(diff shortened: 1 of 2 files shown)


ShruthiPitta:
@sai789 Thank you for the PR! However could you please change the commit message so that it corresponds to the actual issue in JIRA and it would be helpful if the msg reflects the functional issue it causes. It helps in reproducing and reviewing the code too.

ShruthiPitta:
Can you please help in reproducing it in https://product-qa08.mybahmni.org/?

sai789:
Thanks for reviewing.
I have updated BAH-398 Jira card to reproduce the issue. Please have a look at it
I have changed the header of PR as well.

PR #120, opened 2017-12-22
Maharjun, Saikumar | #778 | getting providers who are available for a…

maharjunm:
…ppointments and not changing previous providers who has appointments in past

Maharjun , Saikumar | #778 | Remove console log and fix tests

https://bahmni.atlassian.net/browse/BAH-404

### DIFF
--- a/ui/app/appointments/appointmentConfigInitialization.js
+++ b/ui/app/appointments/appointmentConfigInitialization.js
@@ -42,10 +42,12 @@ angular.module('bahmni.appointments').factory('appointmentConfigInitialization',
                 };
 
                 var getAllProviders = function () {
-                    var params = {v: "custom:(display,person,uuid)"};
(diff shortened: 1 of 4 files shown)


ui/app/appointments/appointmentConfigInitialization.js
```diff
@@ -42,10 +42,12 @@ angular.module('bahmni.appointments').factory('appointmentConfigInitialization',
                 };
 
                 var getAllProviders = function () {
-                    var params = {v: "custom:(display,person,uuid)"};
+                    var params = {v: "custom:(display,person,uuid,attributes:(attributeType:(display),value,voided))"};
```
ShruthiPitta: Instead of filtering on client side, can we fetch the providers based on a certain attribute? I mean do we have a Web API which already does that? If yes can we use it.

PR #121, opened 2018-01-02
BAH-404 |  getting providers who are available for a…

maharjunm:
…ppointments and not changing previous providers who has appointments in past

Maharjun , Saikumar | #778 | Remove console log and fix tests

Maharjun | #778 | Include providers only when they are available for appointments in filters

Maharjun, Saikumar | #778 | should not include retired providers in providers list

https://bahmni.atlassian.net/browse/BAH-404

### DIFF
--- a/ui/app/appointments/constants.js
+++ b/ui/app/appointments/constants.js
@@ -26,7 +26,8 @@ Bahmni.Appointments.Constants = (function () {
         appointmentStatusList: [ "Scheduled", "CheckedIn", "Completed", "Cancelled", "Missed" ],
         regexForTime: /^(?:(?:1[0-2]|0?[1-9]):[0-5]\d\s*[AaPp][Mm])?$/,
         privilegeManageAppointments: 'app:appointments:manageAppointmentsTab',
-        privilegeForAdmin: 'app:appointments:adminTab'
+        privilegeForAdmin: 'app:appointments:adminTab',
+        availableForAppointments: 'Available for appointments'
     };
 })();
 
--- a/ui/app/appointments/controllers/manage/appointmentsCreateController.js
+++ b/ui/app/appointments/controllers/manage/appointmentsCreateController.js
@@ -27,13 +27,13 @@ angular.module('bahmni.appointments')
             };
             var init = function () {
                 wireAutocompleteEvents();
-                var providerCurrentAvailability = "False";
+                var providerCurrentAvailability = false;
                 if (!_.isEmpty(appointmentContext) && !_.isEmpty(appointmentContext.appointment) && !_.isEmpty(appointmentContext.appointment.provider)) {
                     providerCurrentAvailability = isProviderCurrentlyAvailableForAppointments(appointmentContext.appointment.provider, appointmentCreateConfig);
-                }
-                if (_.isEmpty(providerCurrentAvailability)) {
-                    appointmentContext.appointment.provider.person = {display: appointmentContext.appointment.provider.name};
-                    appointmentCreateConfig.providers.push(appointmentContext.appointment.provider);
+                    if (_.isEmpty(providerCurrentAvailability)) {
+                        appointmentContext.appointment.provider.person = {display: appointmentContext.appointment.provider.name};
+                        appointmentCreateConfig.providers.push(appointmentContext.appointment.provider);
+                    }
                 }
                 $scope.appointment = Bahmni.Appointments.AppointmentViewModel.create(appointmentContext.appointment || {appointmentKind: 'Scheduled'}, appointmentCreateConfig);
                 $scope.selectedService = appointmentCreateConfig.selectedService;
--- a/ui/app/appointments/controllers/manage/appointmentsCreateController.js
+++ b/ui/app/appointments/controllers/manage/appointmentsCreateController.js
@@ -19,8 +19,22 @@ angular.module('bahmni.appointments')
             $scope.showStartTimes = [];
             $scope.showEndTimes = [];
 
+            var isProviderCurrentlyAvailableForAppointments = function (selectedProvider, config) {
+                var providers = config.providers;
+                return _.find(providers, function (provider) {
+                    return selectedProvider.uuid === provider.uuid;
+                });
+            };
             var init = function () {
                 wireAutocompleteEvents();
+                var providerCurrentAvailability = "False";
+                if (!_.isEmpty(appointmentContext) && !_.isEmpty(appointmentContext.appointment) && !_.isEmpty(appointmentContext.appointment.provider)) {
+                    providerCurrentAvailability = isProviderCurrentlyAvailableForAppointments(appointmentContext.appointment.provider, appointmentCreateConfig);
+                }
+                if (_.isEmpty(providerCurrentAvailability)) {
+                    appointmentContext.appointment.provider.person = {display: appointmentContext.appointment.provider.name};
+                    appointmentCreateConfig.providers.push(appointmentContext.appointment.provider);
+                }
                 $scope.appointment = Bahmni.Appointments.AppointmentViewModel.create(appointmentContext.appointment || {appointmentKind: 'Scheduled'}, appointmentCreateConfig);
                 $scope.selectedService = appointmentCreateConfig.selectedService;
                 $scope.isPastAppointment = $scope.isEditMode() ? Bahmni.Common.Util.DateUtil.isBeforeDate($scope.appointment.date, moment().startOf('day')) : false;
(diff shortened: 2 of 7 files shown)


mks-d:
@maharjunm is this is the same as PR #123? If yes can we close the current PR?

ShruthiPitta:
@maharjunm Can you please add the actual issue number in commit message?

ui/app/appointments/controllers/manage/appointmentsCreateController.js
```diff
@@ -19,8 +19,22 @@ angular.module('bahmni.appointments')
...
 
+            var isProviderCurrentlyAvailableForAppointments = function (selectedProvider, config) {
+                var providers = config.providers;
+                return _.find(providers, function (provider) {
+                    return selectedProvider.uuid === provider.uuid;
+                });
+            };
             var init = function () {
                 wireAutocompleteEvents();
+                var providerCurrentAvailability = "False";
```
krishnanspace: Hi @maharjunm , on the line  `var providerCurrentAvailability = "False";` why is False written as a String,I mean in quotes?Any specific reason?

krishnanspace: Apart from the above comment,everything looks good so far.  
@mks-d Any comments?

sumanmaity112:
Hi @maharjunm,
Please update the commit message with JIRA card number

djazayeri:
Whoever merges this can use Squash and Merge to set an appropriate commit message (including the ticket number BAH-404)

ui/app/appointments/controllers/manage/appointmentsFilterController.js
```diff
@@ -20,11 +20,13 @@ angular.module('bahmni.appointments')
...
                 }
-                var params = {v: "custom:(display,person,uuid)"};
+                var params = {v: "custom:(display,person,uuid,retired,attributes:(attributeType:(display),value,voided))"};
 
                 spinner.forPromise($q.all([appointmentsServiceService.getAllServicesWithServiceTypes(), providerService.list(params)]).then(function (response) {
-                    $scope.providers = _.uniqBy(response[1].data.results, function (provider) {
-                        return provider.person.display;
+                    $scope.providers = _.filter(response[1].data.results, function (provider) {
+                        return _.find(provider.attributes, function (attribute) {
+                            return attribute.attributeType.display === 'Available for appointments' && attribute.value === true && attribute.voided === false && provider.retired === false;
```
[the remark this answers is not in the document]
maharjunm: I made changes according to feedback. please go through commits and let me know if we need more changes

ui/app/appointments/controllers/manage/appointmentsCreateController.js
```diff
@@ -19,8 +19,22 @@ angular.module('bahmni.appointments')
...
+                return _.find(providers, function (provider) {
+                    return selectedProvider.uuid === provider.uuid;
+                });
+            };
             var init = function () {
                 wireAutocompleteEvents();
+                var providerCurrentAvailability = false;
+                if (!_.isEmpty(appointmentContext) && !_.isEmpty(appointmentContext.appointment) && !_.isEmpty(appointmentContext.appointment.provider)) {
+                    providerCurrentAvailability = isProviderCurrentlyAvailableForAppointments(appointmentContext.appointment.provider, appointmentCreateConfig);
+                    if (_.isEmpty(providerCurrentAvailability)) {
```
angshu: is "_.isEmpty()" the right thing to do? shouldn't we instead do?_.isEqual(providerCurrentlyAvailability, false) be more appropriate, if you must use lodash, rather than conventional equality check.

ui/app/appointments/constants.js
```diff
@@ -26,7 +26,8 @@ Bahmni.Appointments.Constants = (function () {
         appointmentStatusList: [ "Scheduled", "CheckedIn", "Completed", "Cancelled", "Missed" ],
         regexForTime: /^(?:(?:1[0-2]|0?[1-9]):[0-5]\d\s*[AaPp][Mm])?$/,
         privilegeManageAppointments: 'app:appointments:manageAppointmentsTab',
-        privilegeForAdmin: 'app:appointments:adminTab'
+        privilegeForAdmin: 'app:appointments:adminTab',
```
angshu: 1. privileges like "app:appointments:adminTab" seem too UI centric to be defined. Could it be just named "app:appointments:admin" and "app:appointments:manageAppointments"? 
2. how are these privileges created? I don't see them in liquibase migration scripts for the openmrs-module-appointments omod. 
3. In the omod migrations, I  see there are scripts to create "Manage Appointments" and "Manage Appointment Services" etc. Wouldn't these privileges suffice for the UI side as well?

ui/app/appointments/constants.js
```diff
@@ -26,7 +26,8 @@ Bahmni.Appointments.Constants = (function () {
...
+        availableForAppointments: 'Available for appointments'
```
angshu: This is a provide attribute to be defined in OpenMRS. Does the appointments omod create this attribute as part of migration script?

ui/app/appointments/constants.js
pramidatumma: 1. These privileges are already existing in the production. If we change the name of these privileges, we need to write migrations to take care of the existing users/privileges. As this is not part of this issue, we can create a tech debt.
2. These privileges are created using liquibase migration scripts in bahmni core omod
3. These privileges are driven more by UI design

ui/app/appointments/constants.js
pramidatumma: Yes

ui/app/appointments/controllers/manage/appointmentsCreateController.js
```diff
@@ -19,8 +19,22 @@ angular.module('bahmni.appointments')
...
+                    return selectedProvider.uuid === provider.uuid;
+                });
+            };
             var init = function () {
                 wireAutocompleteEvents();
+                var providerCurrentAvailability = "False";
+                if (!_.isEmpty(appointmentContext) && !_.isEmpty(appointmentContext.appointment) && !_.isEmpty(appointmentContext.appointment.provider)) {
+                    providerCurrentAvailability = isProviderCurrentlyAvailableForAppointments(appointmentContext.appointment.provider, appointmentCreateConfig);
+                }
+                if (_.isEmpty(providerCurrentAvailability)) {
```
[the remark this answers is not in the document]
sai789: We have made changes for this. Changes are for isEmpty() replacement

ui/app/appointments/constants.js
angshu: 1. Yes. If you are changing or introducing privileges, then the module migrations must take care of it, irrespective of whether the changes are relevant to the UI or not. I am not sure about another card for this (outside of 0.91 release), as it would then compel others to do this manually. 
2. For the privileges themselves - I definitely find them overlapping. Any specific reason to have ‘app:appointments:manageAppointments’ while there is already one ‘Manage Appointments’, similarly there exists one for admin  (irrespective of whether the admin part is from UI). Its preferable if we can have one set of privileges which could be applied to both UI and backend layers.  Else, it makes it confusing and difficult to maintain. Also the service calls must be guarded by the same set of privileges. We would look to introduce ui specific privileges if its a broader level grouping at the feature (which can be also done using roles), and if we can't leverage core module privileges (or they don't make sense).

ui/app/appointments/controllers/manage/appointmentsFilterController.js
angshu: The comparison still reflects the string 'Available for appointments' 
> 
Also, its preferred that you put the guard statements ahead. for example
>  attribute.voided === false && attribute.value === true && provider.retired === false && attribute.attributeType.display === 'Available for appointments';

This way the expression does not need to be evaluated to the completely.

maharjunm:
@angshu I pushed the code according to the feedback. please let me know if you need more changes

maharjunm:
@angshu I changed the code according to your feedback and pushed. Review it and let me know If we need more changes :)

ui/test/unit/appointments/controllers/manage/appointmentsCreateController.spec.js
```diff
@@ -913,5 +913,30 @@ describe("AppointmentsCreateController", function () {
...
         $scope.onKeyDownOnEndTime();
         expect($scope.showEndTimes).toEqual($scope.endTimes);
-    })
+    });
+
+    it('should include appointment provider if the provider is not available in all available providers', function () {
+        var allAvailableProviders = [{name: 'superman', uuid: '1'}, {name: 'mahmoud_h', uuid: '2'}];
+        appointmentContext.appointment = {provider: {name: 'someone', uuid: '3'}};
+        appointmentCreateConfig.providers = allAvailableProviders;
+
```
krishnanspace: Can you please remove the blank lines?Apart from this,everything looks fine to me

maharjunm:
@krishnanspace  I updated according to feedback. review it :)

PR #129, opened 2018-01-29
Suman | BAH-447 | Introduce 'hideEmptyRecords' config key to hide emp…

sumanmaity112:
…ty records from obsToObsFlowSheet display control

### DIFF
--- a/ui/app/common/displaycontrols/tabularview/directives/obsToObsFlowSheet.js
+++ b/ui/app/common/displaycontrols/tabularview/directives/obsToObsFlowSheet.js
@@ -23,6 +23,15 @@ angular.module('bahmni.common.displaycontrol.obsVsObsFlowSheet').directive('obsT
                 });
             };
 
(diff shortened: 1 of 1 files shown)


sumanmaity112:
@mks-d there is some unit test for this directive, but already there is no unit test for the method `getObsInFlowSheet`. All the method are already mocked in `beforeEach`, i.e I didn't write any.

ui/app/common/displaycontrols/tabularview/directives/obsToObsFlowSheet.js
```diff
@@ -23,6 +23,15 @@ angular.module('bahmni.common.displaycontrol.obsVsObsFlowSheet').directive('obsT
                 });
             };
 
+            const removeEmptyRecords = function (records) {
```
djazayeri: personally I think it's bad style, and can be confusing for the caller, if you both mutate the input _and_ also return it. (But I will still merge anyway.)

PR #133, opened 2018-02-15
Show Units irrespective of ranges configured

drkreddy:
In Investigation chart display control, Test units are displayed only when ranges are configured. 
Made changes to show units when configured irrespective of ranges.


ui/app/clinical/displaycontrols/investigationresults/models/tabularLabOrderResults.js
```diff
@@ -50,6 +50,10 @@ Bahmni.Clinical.TabularLabOrderResults = (function () {
             return testOrderLabel.minNormal && testOrderLabel.maxNormal;
         };
 
+        this.hasUnits = function (testOrderLabel) {
```
sumanmaity112: @drkreddy 
Can you use *_.isEmpty*

binduak: @drkreddy Can you please add test case for this.

PR #138, opened 2018-03-09
BAH-482|Improve error message for invalid start/end times

divyabaid16:
[Improve error message for invalid start/end times](https://bahmni.atlassian.net/browse/BAH-482?src=confmacro)


ShruthiPitta:
@divyabaid16 Thank you for improving the error message. Usual convention is to Add the issue number from jira in the commit message.

PR #149, opened 2018-04-17
BAH-497: Prompt to save if home or search clicked in registration screen

krishnanspace:
1.A confirm box has been added when ever the user tries to navigate away from the patient registration screen.  
2.The above prompt is enabled through the following config being set to true in registration/app.js: "showSaveConfirmDialog" : true
3. However "stateChangeStart" and "locationChangeStart" are not getting triggered for the Home button and preventing to leverage the above functionality.
4. The desired outcome is achievable by setting an "id" attribute to the Home button and calling a function when the button is clicked.
5.Navigating away through

- Clicking on Search button
- URL  change
- Navigating away from patient registration page
- Clicking the Chrome's back button triggers the confirmation prompt to appear as it does in 'clinical'.


sumanmaity112:
Hi @krishnanspace,
can you please squash these two commits into one. Also, fix the codeclimate issues

PR #153, opened 2018-04-22
BAH-433 Filters are not getting reapplied, when user change the appointment status

smartyrad:
BAH-433 Filters are not getting reapplied, when user change the appointment status

### DIFF
--- a/ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
+++ b/ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
@@ -168,6 +168,15 @@ angular.module('bahmni.appointments')
                 showPopUp(scope);
             };
 
+            $scope.getAppointmentsForDateNew = function (viewDate) {
+                $stateParams.viewDate = viewDate;
+                var params = {forDate: viewDate};
+                spinner.forPromise(appointmentsService.getAllAppointments(params).then(function (response) {
+                    $scope.appointments = response.data;
+                    $scope.filteredAppointments = appointmentsFilter($scope.appointments, $stateParams.filterParams);
+                }));
+            };
+
             var changeStatus = function (toStatus, onDate, closeConfirmBox) {
                 var message = $translate.instant('APPOINTMENT_STATUS_CHANGE_SUCCESS_MESSAGE', {
                     toStatus: toStatus
@@ -177,6 +186,7 @@ angular.module('bahmni.appointments')
                     $scope.selectedAppointment.status = response.data.status;
                     closeConfirmBox();
                     messagingService.showMessage('info', message);
+                    $scope.getAppointmentsForDateNew($stateParams.viewDate);
                 });
             };
 
@@ -199,6 +209,7 @@ angular.module('bahmni.appointments')
             var showPopUp = function (popUpScope) {
                 popUpScope.no = function (closeConfirmBox) {
                     closeConfirmBox();
+                    console.log($stateParams.viewDate);
                 };
                 confirmBox({
                     scope: popUpScope,


ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
```diff
@@ -199,6 +209,7 @@ angular.module('bahmni.appointments')
             var showPopUp = function (popUpScope) {
                 popUpScope.no = function (closeConfirmBox) {
                     closeConfirmBox();
+                    console.log($stateParams.viewDate);
```
sumanmaity112: Hi @smartyrad,
Please remove `console.log`

smartyrad: Hi @sumanmaity112 , sure. But this is not the final PR. I needed some help in reviewing the code to decide the course of action to be taken to fix this

smartyrad:
@mks-d @sumanmaity112  I am facing a small issue with respect to this issue. My current implementation is not final and if not anything else, it does require me to change the name of the new function that I have defined. However, the main issue is that for fixing this, I have called the "getAppointmentsForDate" function.

Now,
1) If I just call the function then, [this](https://github.com/Bahmni/openmrs-module-bahmniapps/blob/8c4c870dbe4059528963ab5904dd97aabf07027a/ui/test/unit/appointments/controllers/manage/list/appointmentsListViewController.spec.js#L869) test fails and the reason for it failing is because of the $scope.selectedappointment variable being set to undefined within the getAppointmentsForDate function as can be seen [here](https://github.com/Bahmni/openmrs-module-bahmniapps/blob/4e3f8376ce7caf9ce5e213de13d94c4715718c63/ui/app/appointments/controllers/manage/list/appointmentsListViewController.js#L41)

Soln 1: Remove that line which sets it to undefined
Soln 2: Create a new function that does not set it to undefined (increases redundancy as the function has very few changes from the source function)[ THIS is what I have done in this commit]
Soln 3: Remove the particular check from the test

As a side note, I would like to mention that the corresponding function in the Calendar view has the same function defined but without this line which sets the value to undefined as can be seen [here](https://github.com/Bahmni/openmrs-module-bahmniapps/blob/4e3f8376ce7caf9ce5e213de13d94c4715718c63/ui/app/appointments/controllers/manage/calendar/appointmentsCalendarViewController.js#L81)

Kindly suggest as to what we could do and if you have any other thoughts with the fundamental approach itself then kindly do let me know :)

smartyrad:
Also kindly ignore the codeclimate build error for now :)

ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
djazayeri: @smartyrad we are looking to do the 0.91 release very soon. Does this PR solve the problem, or not yet?

smartyrad: @djazayeri Yes, this issue is almost fixed. But due to a particular test failing, I had listed down a few steps that I could possibly take to solve the issue. Can you please help me out with this?

djazayeri:
I spent a few minutes looking at this, but it's the end of the week and my brain is a bit fried. I'll try to at least say _something_.

You're saying that the current behavior of getAppointmentsForDate in the list view unselects the current appointment. I can imagine this might be happening to ensure that if you change the date, we don't end up with an invalid selection that doesn't match the filter. (That said, if the behavior is different in the calendar view I would assume that one of them is wrong, I just don't know which one offhand.)

Also, when one changes the status of an appointment, this should _not_ deselect the appointment (at least I don't think the user would want that), so that test is "failing correctly".

Other possible options:
* add a parameter to that function, which determines whether the selection will remain
* write some more sophisticated logic to only unselect the appointment if it's an invalid selection
* remove the line that unselects, but change the current calling code to unselect it after calling getAppointmentsForDate

ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
```diff
@@ -177,6 +186,7 @@ angular.module('bahmni.appointments')
                     $scope.selectedAppointment.status = response.data.status;
                     closeConfirmBox();
                     messagingService.showMessage('info', message);
+                    $scope.getAppointmentsForDateNew($stateParams.viewDate);
```
pramidatumma: Instead of getting new appointments, you can just update the scope variable filteredAppointments

PR #156, opened 2018-04-26
Bah 446: Form is not saved when concept of Image type is marked as Mandatory 

krishnanspace:
Just had to make 1 modification to get the Form to save when an Image is set to Mandatory.
Here is the JIRA Card:
https://bahmni.atlassian.net/browse/BAH-446

The reason why the problem was occurring is mentioned in the JIRA Card

### DIFF
--- a/ui/app/common/concept-set/directives/imageUpload.js
+++ b/ui/app/common/concept-set/directives/imageUpload.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module('bahmni.common.conceptSet')
-    .directive('imageUpload', ['visitDocumentService', 'messagingService', 'spinner', '$rootScope', function (visitDocumentService, messagingService, spinner, $rootScope) {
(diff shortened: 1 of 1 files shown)


krishnanspace:
@mks-d Actually its not an issue per se. When you upload 1 image, one more image upload concept(a clone) is created automatically.
If you upload a second image , one more clone is created.
Now all these clones are marked as mandatory.Thats why earlier we were not able to save the form as the clone was creating problems.
But from the code's perspective, everything is working as intended
So I dont know whether a unit test would be applicable
Also about the travis build, its showing that test related to OT are failing, no idea why its doing so. Its even failing for others too I think

krishnanspace:
@mks-d Can the Travis CI be executed again ? It was failing the last time because of the OT Tests failing

mks-d:
@krishnanspace:

image


image

krishnanspace:
image
@mks-d I dont see the option to restart the build

mks-d:
@krishnanspace I just restarted it then:
https://travis-ci.org/Bahmni/openmrs-module-bahmniapps/builds/371390706

PR #158, opened 2018-05-08
BAH-526 | Fixed issue of creating duplicate appointments

maharjunm:
https://bahmni.atlassian.net/browse/BAH-526


djazayeri:
@maharjunm I see edits to existing unit tests but I don't see any unit tests. Given that the ticket gives three different scenarios to reproduce, my baseline expectation is at least 3 new unit tests.

maharjunm:
@djazayeri Above test cases will cover most of the scenarios. If I need to add more test cases please let me know :)

PR #159, opened 2018-05-09
BAH-528 | fixed error while deselecting concept in microbiology resul…

maharjunm:
…t form

https://bahmni.atlassian.net/browse/BAH-528

### DIFF
--- a/ui/app/common/concept-set/directives/conceptSet.js
+++ b/ui/app/common/concept-set/directives/conceptSet.js
@@ -238,6 +238,7 @@ angular.module('bahmni.common.conceptSet')
                 };
 
                 var runFormConditionForAllObsRecursively = function (formName, rootObservation) {
+                    rootObservation && rootObservation.groupMembers &&
                     rootObservation.groupMembers.forEach(function (observation) {
                         var conditionFn = Bahmni.ConceptSet.FormConditions.rules && Bahmni.ConceptSet.FormConditions.rules[observation.concept.name];
                         if (conditionFn != null) {
--- a/ui/app/common/concept-set/directives/conceptSet.js
+++ b/ui/app/common/concept-set/directives/conceptSet.js
@@ -238,7 +238,7 @@ angular.module('bahmni.common.conceptSet')
                 };
 
                 var runFormConditionForAllObsRecursively = function (formName, rootObservation) {
-                    _.each(rootObservation.groupMembers, function (observation) {
+                    rootObservation.groupMembers.forEach(function (observation) {
                         var conditionFn = Bahmni.ConceptSet.FormConditions.rules && Bahmni.ConceptSet.FormConditions.rules[observation.concept.name];
                         if (conditionFn != null) {
                             var flattenedObs = ObservationUtil.flattenObsToArray([rootObservation]);


ui/app/common/concept-set/directives/conceptSet.js
```diff
@@ -238,7 +238,7 @@ angular.module('bahmni.common.conceptSet')
                 };
 
                 var runFormConditionForAllObsRecursively = function (formName, rootObservation) {
-                    _.each(rootObservation.groupMembers, function (observation) {
+                    rootObservation.groupMembers.forEach(function (observation) {
```
sumanmaity112: Hi @maharjunm, this fix will fail if `rootObservation.groupMembers` is `undefined`. Afterthat `_.each` is a `null` safe operation which internally uses the `forEach` method.

djazayeri: I also don't see how this change would fix anything. And there's no unit test.

sai789: We made a commit for undefined checks for `rootObservation` and `rootObservation.groupMembers`.

`_.each` does not adhere to concurrent modification of iterating list. Using `forEach` that works for concurrent modification

sravanthi17: @sai789 Can you also add tests for this ?

ui/app/common/concept-set/directives/conceptSet.js
```diff
@@ -238,7 +238,8 @@ angular.module('bahmni.common.conceptSet')
                 };
 
                 var runFormConditionForAllObsRecursively = function (formName, rootObservation) {
-                    _.each(rootObservation.groupMembers, function (observation) {
+                    rootObservation && rootObservation.groupMembers &&
+                    rootObservation.groupMembers.forEach(function (observation) {
```
sumanmaity112: Hi @maharjunm, please add UT for these changes

ui/app/common/concept-set/directives/conceptSet.js
```diff
@@ -238,7 +238,8 @@ angular.module('bahmni.common.conceptSet')
                 };
 
                 var runFormConditionForAllObsRecursively = function (formName, rootObservation) {
-                    _.each(rootObservation.groupMembers, function (observation) {
+                    rootObservation && rootObservation.groupMembers &&
```
sumanmaity112: Hi @maharjunm, Please change `rootObservation && rootObservation.groupMembers && rootObservation.groupMembers.forEach(function(){
....
)` to `rootObservation && _.each(rootObservation.groupMembers, function(){
....
})`

sai789: Hi @sumanmaity112 
This itself is the fix. We explained in previous comments why we are migrating from _.each to .forEach.

sumanmaity112: Hi @sai789, 
If I understood your comments, you need to check * rootObservation*  is `undefined` or not.
You can do that using the following way as well
`rootObservation && _.each(rootObservation.groupMembers, function(){ .... })`. This same as `rootObservation && rootObservation.groupMembers && rootObservation.groupMembers.forEach(function(){ .... )`

sai789: Hi @sumanmaity112, 
We have replaced _.each with forEach because  _.each does not adhere to concurrent modification of iterating list. Using forEach that works for concurrent modification.
This change is the fix

PR #163, opened 2018-05-15
BAH-415 Mandatory Boolean Concept on Visit Page 

Pasayadaan:
Added condition for Mandatory Boolean Concept on Visit Page as explained
https://talk.openmrs.org/t/error-while-saving-visit-page-with-mandatory-boolean-concept/14967

### DIFF
--- a/ui/app/registration/controllers/visitController.js
+++ b/ui/app/registration/controllers/visitController.js
@@ -230,7 +230,7 @@ angular.module('bahmni.registration')
             var isValid = function (mandatoryConcepts) {
                 var concept = mandatoryConcepts.filter(function (mandatoryConcept) {
                     if ((mandatoryConcept.isNumeric() && mandatoryConcept.value === 0) ||
-                        (mandatoryConcept.isBoolean && mandatoryConcept.value != null)) {
+                        (mandatoryConcept.isBoolean && mandatoryConcept.hasValue())) {
                         return false;
                     }
                     if (mandatoryConcept instanceof Bahmni.ConceptSet.Observation &&
--- a/ui/app/registration/controllers/visitController.js
+++ b/ui/app/registration/controllers/visitController.js
@@ -229,7 +229,8 @@ angular.module('bahmni.registration')
             };
             var isValid = function (mandatoryConcepts) {
                 var concept = mandatoryConcepts.filter(function (mandatoryConcept) {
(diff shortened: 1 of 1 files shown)


Pasayadaan:
Ok. Let me test hasvalue() for other data types too and make the necessary change

sumanmaity112:
Cool :)

Pasayadaan:
While hasValue() can handle numeric, boolean, string the special treatment done in this function for [multiselect](https://github.com/Bahmni/openmrs-module-bahmniapps/blob/master/ui/app/registration/controllers/visitController.js#L235-L241) will have to be retained.

sumanmaity112:
@Pasayadaan, can you please check with the latest changes (`hasValue`). I feel everything will work fine

Pasayadaan:
Yes, we have checked the latest hasValue(). But for mandatory multiselect the mandatory [mandatoryConceptGroup](https://github.com/Bahmni/openmrs-module-bahmniapps/blob/master/ui/app/registration/controllers/visitController.js#L209) holds 2 concept nodes one is of instance multiselectobservation and the other is instance of observation. (I don't know why it has 2 nodes for one concept). For this the special handling of multiselect is required. It may be a bug somewhere else that it populates 2 nodes for one concept.

Pasayadaan:
For a Multiselect mandatory concept when 2 values are selected it, in isValid function gets an array of 4 objects as – one is Multiselect defined in conceptSetUI, 2 conceptset.observation for the 2 selected values and there is fourth empty conceptset.observation for which hasValue fails. Hence that additional check for multiselect cannot be removed.
If that additional check for multiselect is acceptable for the code review at this time, then the necessary changes can be done to keep hasValue () for other data types.

`var concept = mandatoryConcepts.filter(function (mandatoryConcept) {
                    if (mandatoryConcept.hasValue()) {
                        return false;
                    }
                    if (mandatoryConcept instanceof Bahmni.ConceptSet.Observation &&
                        mandatoryConcept.conceptUIConfig && mandatoryConcept.conceptUIConfig.multiSelect) {
                        return false;
                    }
                    if (mandatoryConcept.isMultiSelect) {
                        return _.isEmpty(mandatoryConcept.getValues());
                    }
                    return !mandatoryConcept.value;
                });
                return _.isEmpty(concept);`

PR #165, opened 2018-05-16
https://bahmni.atlassian.net/browse/BAH-465

shilpa-iplit:
Read all Patient Document set members and displayed it in drop down as document file type.
Code was changed for populating this drop down and after selection change setting appropriate selected file type with uploaded file.

### DIFF
--- a/ui/app/document-upload/controllers/documentController.js
+++ b/ui/app/document-upload/controllers/documentController.js
@@ -15,7 +15,7 @@ angular.module('opd.documentupload')
             var locationUuid = sessionService.getLoginLocationUuid();
 
             $scope.visits = [];
-            $scope.allConcepts = [];
+            $scope.fileTypeConcepts = [];
             $scope.toggleGallery = true;
             $scope.conceptNameInvalid = false;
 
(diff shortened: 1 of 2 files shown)


ui/app/document-upload/controllers/documentController.js
```diff
@@ -142,9 +143,21 @@ angular.module('opd.documentupload')
                         v: "custom:(uuid,setMembers:(uuid,name:(name)))"
                     }
                 }).then(function (response) {
+                    if (response.data.results[0].setMembers && response.data.results[0].setMembers.length > 0) {
+                        response.data.results[0].setMembers.forEach(function (concept) {
+                            var conceptToAdd = {
+                                'concept': {
+                                    uuid: concept.uuid,
+                                    name: concept.name.name,
+                                    editableName: concept.name.name
```
SwathiVarkala: why editableName on fileType Concept? can we remove this?
below suggested approach for fileType selection, editableName will not be used.

ui/app/document-upload/controllers/documentController.js
```diff
@@ -142,9 +143,21 @@ angular.module('opd.documentupload')
...
+                                }
+                            };
+                            $scope.allConcepts.push(conceptToAdd);
+                        });
+                    }
                     var topLevelConcept = response.data.results[0];
                     topLevelConceptUuid = topLevelConcept ? topLevelConcept.uuid : null;
-                    setDefaultConcept(topLevelConcept);
+                    // setDefaultConcept(topLevelConcept);
```
SwathiVarkala: - donot comment but remove the lines of old code to be removed as part this change
- if setDefaultConcept is not used, it can be removed.

ui/app/document-upload/controllers/documentController.js
```diff
@@ -15,6 +15,7 @@ angular.module('opd.documentupload')
             var locationUuid = sessionService.getLoginLocationUuid();
 
             $scope.visits = [];
+            $scope.allConcepts = [];
```
SwathiVarkala: We can name it better.
Ex: allConcepts to fileTypeConcepts, filteredData to selectedConcept

SwathiVarkala:
As per the card, the ask is to have some fileType populated by default in dropdown. Is this taken care in this PR?

ui/app/document-upload/controllers/documentController.js
```diff
@@ -142,9 +143,21 @@ angular.module('opd.documentupload')
                         v: "custom:(uuid,setMembers:(uuid,name:(name)))"
                     }
                 }).then(function (response) {
+                    if (response.data.results[0].setMembers && response.data.results[0].setMembers.length > 0) {
```
SwathiVarkala: Please ensure do empty check. If there is no TopLevelConcept then here the application will enter into exception state.
TypeError: Cannot read property 'setMembers' of undefined
    at https://192.168.33.10/bahmni/document-upload/controllers/documentController.js:146:50

SwathiVarkala:
Please do single data binding where ever applicable to reduce no of angular watches degrading performance.

shilpa-iplit:
These are the PR review comments addressed so far:
- cleanup when ever possible
- naming conventions
- avoid loops to have better performance
- first file type option populated by default in drop down.

These are the comments which need not be addressed:
- Why editableName on file type Concept? -> its needed to get the user modified item in drop down
- Please ensure do empty check. If there is no TopLevelConcept then here the application will enter into exception state. -> No longer the case
- Please do single data binding where ever applicable.

PR #179, opened 2018-06-19
BAH-497: Prompt to save if home or search clicked in registration screen

krishnanspace:
Solved the problem of the Prompt appearing on "Enter Visit Details" page

### DIFF
--- a/ui/app/registration/controllers/patientCommonController.js
+++ b/ui/app/registration/controllers/patientCommonController.js
@@ -23,7 +23,7 @@ angular.module('bahmni.registration')
             var isHref = false;
 
             $rootScope.onHomeNavigate = function (event) {
-                if ($scope.showSaveConfirmDialogConfig) {
+                if ($scope.showSaveConfirmDialogConfig && $state.current.name != "patient.visit") {
                     event.preventDefault();
                     $scope.targetUrl = event.currentTarget.getAttribute('href');
                     isHref = true;
--- a/ui/test/unit/registration/controllers/patientCommonController.spec.js
+++ b/ui/test/unit/registration/controllers/patientCommonController.spec.js
@@ -2,7 +2,7 @@
 
 describe('PatientCommonController', function () {
 
-    var $aController, $httpBackend, scope, appService, rootScope, patientAttributeService;
+    var $aController, $httpBackend, scope, appService, rootScope, patientAttributeService, $state;
     var spinner = jasmine.createSpyObj('spinner', ['forPromise']);
     var $compile;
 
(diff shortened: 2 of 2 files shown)


ui/app/registration/controllers/patientCommonController.js
```diff
@@ -23,7 +23,7 @@ angular.module('bahmni.registration')
             var isHref = false;
 
             $rootScope.onHomeNavigate = function (event) {
-                if ($scope.showSaveConfirmDialogConfig) {
+                if ($scope.showSaveConfirmDialogConfig && $state.current.name != "patient.visit") {
```
mks-d: @krishnanspace is there no unit test to back this behaviour?

krishnanspace:
@mks-d Added  the required test case

ui/test/unit/registration/controllers/patientCommonController.spec.js
```diff
@@ -76,6 +76,20 @@ describe('PatientCommonController', function () {
...
+        scope.onHomeNavigate = jasmine.createSpy("onHomeNavigate");
+        scope.confirmationPrompt = jasmine.createSpy("confirmationPrompt");
+        var element = angular.element("<a ng-click='onHomeNavigate()'>");
+        var compiled = $compile(element)(scope);
+        compiled.triggerHandler('click');
+        expect(scope.onHomeNavigate).toHaveBeenCalled();
+        expect(scope.showSaveConfirmDialogConfig).toBe(true);
+        expect($state.current.name).toBe("patient.visit");
+        expect(scope.confirmationPrompt).not.toHaveBeenCalled();
+    });
```
mks-d: OK that's great but can you also demonstrate that the popup does **not** show up if `$state.current.name` is not `"patient.visit"`? Because that is precisely the behaviour that was spotted during QA.

mks-d: @krishnanspace I meant that we would need one specific test case method that shows that the popup is not activated when `$state.current.name` is not `"patient.visit"`. Does that make sense?

I'm asking this because that's the exact bug that was spotted during QA.

krishnanspace: @mks-d It will show the pop-up for all other states except patient.visit but wont show it on patient.visit state. Earlier it was found out that the prompt was appearing in patient.visit too.

krishnanspace: I mean the pop-up would be activated when the $state.current.name is not patient.visit

angshu: Please note that state.current.name is set to patient.visit only for the visit details (registration second page). From the card (BAH-497), its understood that the confirmation is to be shown only for the first page (patient edit page). The code looks ok for that.

PR #182, opened 2018-07-25
date picker makes a single API call on change and on blur. Also delete and…

shilpa-iplit:
… backspace will make a single API call on blur


shilpa-iplit:
https://bahmni.atlassian.net/browse/BAH-435  PR for this card.

PR #198, opened 2018-09-06
Auto refresh of Appointments

rajur7:
[JIRA link for Auto Refresh Story](https://bahmni.atlassian.net/browse/BAH-624)

Configuration:- 
    We need to add a configuration to enable this feature. We have to configure with a time interval by providing value to `autoRefreshIntervalInSeconds` in ***openmrs/apps/appointments/app.json*** in config.

### DIFF
--- a/ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
+++ b/ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
@@ -45,27 +45,28 @@ angular.module('bahmni.appointments')
                     $scope.filteredAppointments = appointmentsFilter($scope.appointments, $stateParams.filterParams);
                     $rootScope.appointmentsData = $scope.filteredAppointments;
                     autoRefreshStatus = true;
-                })
+                });
             };
 
-            var autoRefreshInterval = function (){
-                if(!enableAutoRefresh){
+            var autoRefresh = (function () {
+                if (!enableAutoRefresh) {
                     return;
                 }
                 var autoRefreshIntervalInMilliSeconds = appService.getAppDescriptor().getConfigValue('autoRefreshIntervalInMilliSeconds');
                 return $interval(function () {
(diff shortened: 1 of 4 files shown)


ui/app/appointments/controllers/manage/list/appointmentsListViewController.js
```diff
@@ -36,6 +38,53 @@ angular.module('bahmni.appointments')
                 $scope.isFilterOpen = $stateParams.isFilterOpen;
             };
 
+            var autoRefreshStatus = true;
```
sumanmaity112: @rajur7, Please move this ``var autoRefreshStatus = true;`` to top as this is the default value

ui/test/unit/appointments/controllers/manage/calendar/appointmentsCalendarViewController.spec.js
```diff
@@ -474,5 +480,65 @@ describe('AppointmentsCalendarViewController', function () {
...
+
+    it('should call interval function when autoRefreshIntervalInSeconds is defined', function () {
+        appDescriptor.getConfigValue.and.callFake(function (value) {
+            if (value === 'autoRefreshIntervalInSeconds') {
+                return 10;
+            }
+            return undefined;
+        });
+
+        createController();
```
sumanmaity112: @rajur7, If I am not wrong, you can pull `createController()` to `beforeEach` method

sowmika148: @sumanmaity112 , we are calling createController method to initialise the private variables required for these tests.

sumanmaity112: @sowmika148, then would it be `beforeEach` thing?

PR #201, opened 2018-09-25
Shilpa|BAH-513|deleted Obs form can be readded

shilpa-iplit:
Original template form's isAdded is made false so that it could be re-added again.
Also After second consecutive deletion and then addition for third time was causing issue in form addition in function addTemplate as code had that form already added. So fixed it.


SwathiVarkala:
@shilpa-iplit , weren't there any tests cases that cover scenarios around adding and removing forms?

PR #203, opened 2018-10-01
Auto Refresh of Appointments

rajur7:
[JIRA link for Auto Refresh Story](https://bahmni.atlassian.net/browse/BAH-624)

Configuration:- 
    We need to add a configuration to enable this feature. We have to configure with a time interval by providing value to `autoRefreshIntervalInSeconds` in ***openmrs/apps/appointments/app.json*** in config.

### DIFF
--- a/ui/app/appointments/controllers/manage/calendar/appointmentsCalendarViewController.js
+++ b/ui/app/appointments/controllers/manage/calendar/appointmentsCalendarViewController.js
@@ -1,8 +1,9 @@
 'use strict';
 
 angular.module('bahmni.appointments')
-    .controller('AppointmentsCalendarViewController', ['$scope', '$state', '$translate', 'spinner', 'appointmentsService', 'appointmentsFilter', '$rootScope',
-        function ($scope, $state, $translate, spinner, appointmentsService, appointmentsFilter, $rootScope) {
+    .controller('AppointmentsCalendarViewController', ['$scope', '$state', '$translate', 'spinner', 'appointmentsService', 'appointmentsFilter', '$rootScope', '$interval', 'appService',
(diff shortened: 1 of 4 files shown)


ui/app/appointments/controllers/manage/calendar/appointmentsCalendarViewController.js
```diff
@@ -1,8 +1,12 @@
 'use strict';
 
 angular.module('bahmni.appointments')
-    .controller('AppointmentsCalendarViewController', ['$scope', '$state', '$translate', 'spinner', 'appointmentsService', 'appointmentsFilter', '$rootScope',
-        function ($scope, $state, $translate, spinner, appointmentsService, appointmentsFilter, $rootScope) {
+    .controller('AppointmentsCalendarViewController', ['$scope', '$state', '$translate', 'spinner', 'appointmentsService', 'appointmentsFilter', '$rootScope', '$interval', 'appService',
+        function ($scope, $state, $translate, spinner, appointmentsService, appointmentsFilter, $rootScope, $interval, appService) {
+            var autoRefreshIntervalInSeconds = parseInt(appService.getAppDescriptor().getConfigValue('autoRefreshIntervalInSeconds'));
```
krishnanspace: I think you should also add a check to accept only positive numbers or if its negative(an unlikely situation), convert the negative number to positive

rajur7: By default negative numbers will be handled in the interval function. If we give any negative number then the setInterval function will  automatically set 4ms time as interval time which is minimum default value

krishnanspace: > By default negative numbers will be handled in the interval function. If we give any negative number then the setInterval function will automatically set 4ms time as interval time which is minimum default value

Ok  cool. Everything is fine then

PR #205, opened 2018-10-11
Differentiate multiple appointments 

sowmika148:
Differentiate multiple appointments created in same slot for same patient in calendar view popup
[JIRA CARD](https://bahmni.atlassian.net/browse/BAH-670)


nehagupta2123:
BAH 670 [Appt. Scheduling]Unable to differentiate between Multiple appointments for same time slot and same provider.
https://bahmni.atlassian.net/browse/BAH-670

PR #216, opened 2019-01-02
Fix camera issue with latest version of chrome

ivange94:
URL.createObjectURL() which we used to access the webcam
was deprecated in Chrome versions
prior to Chrome 71. In Chrome 71, this method was removed
and as a result a JavaScript error was thrown everytime you
try to access the Camera.

Resolves: #738

### DIFF
--- a/ui/app/common/photo-capture/directives/capturePhoto.js
+++ b/ui/app/common/photo-capture/directives/capturePhoto.js
@@ -72,7 +72,7 @@ angular.module('bahmni.common.photoCapture')
                     navigator.getUserMedia(
                         {video: true, audio: false},
                         function (localMediaStream) {
-                            captureVideo.src = $window.URL.createObjectURL(localMediaStream);
+                            captureVideo.srcObject = localMediaStream;
                             captureActiveStream = localMediaStream;
                             captureDialogElement.dialog('open');
                         },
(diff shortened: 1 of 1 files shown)


ui/app/common/photo-capture/directives/capturePhoto.js
```diff
@@ -72,7 +72,7 @@ angular.module('bahmni.common.photoCapture')
                     navigator.getUserMedia(
                         {video: true, audio: false},
                         function (localMediaStream) {
-                            captureVideo.src = $window.URL.createObjectURL(localMediaStream);
+                            captureVideo.srcObject = localMediaStream;
```
angshu: I will probably be a bit more defensive so that just in case someone is using an older browser. Instead of just updating the source like that, I would put a check if "navigator.mediaDevices" is defined and then use it in a if block preceding the existing code. like in talk thread [here](https://talk.openmrs.org/t/camera-not-working-during-registration/21132/8) 
Also this way, we can use the suggested api for "navigator.mediaDevices", if not then fall back to the older navigator.getUserMedia (existing code)

angshu: @ivange94 you are right, and Bahmni does recommend using Chrome. Bahmni is often used on-premise and people may not have internet and therefore automatic updates. We have known it to be used in tabs/ipads and safari and FF.  I am probably just being defensive

PR #224, opened 2019-01-23
Sowmika | BAH-747 | Set current date as default for date type person attribute 

nehagupta2123:
https://bahmni.atlassian.net/browse/BAH-747

### DIFF
--- a/ui/app/registration/controllers/createPatientController.js
+++ b/ui/app/registration/controllers/createPatientController.js
@@ -51,9 +51,32 @@ angular.module('bahmni.registration')
                     }).value();
                 };
 
-                _.chain(personAttributeTypes)
+                var isDateTypeAndDefaultValueToday = function (personAttributeType) {
+                    return isDateType(personAttributeType) && isDefaultValueToday(personAttributeType);
+                };
+
+                var isDateType = function (personAttributeType) {
+                    return personAttributeType.format === "org.openmrs.util.AttributableDate";
+                };
+
+                var isDefaultValueToday = function (personAttributeType) {
+                    if (defaults[personAttributeType.name].toLowerCase() === "today") {
+                        return true;
+                    }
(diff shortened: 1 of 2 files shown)


ui/app/registration/controllers/createPatientController.js
```diff
@@ -51,9 +51,32 @@ angular.module('bahmni.registration')
...
+
+                var isDateType = function (personAttributeType) {
+                    return personAttributeType.format === "org.openmrs.util.AttributableDate";
+                };
+
+                var isDefaultValueToday = function (personAttributeType) {
+                    if (defaults[personAttributeType.name].toLowerCase() === "today") {
+                        return true;
+                    }
+                    $scope.patient[personAttributeType.name] = '';
```
sravya-kanagarla: @sowmika148 Instead of setting empty value in boolean method, we can set it in 'setDefaultTodayDate' method which we have separately to set data to the patient. Please refactor accordingly.

ui/app/registration/controllers/createPatientController.js
```diff
@@ -51,9 +51,32 @@ angular.module('bahmni.registration')
...
+                    $scope.patient[personAttributeType.name] = new Date();
+                };
+
+                var defaultsWithAnswers = _.chain(personAttributeTypes)
                     .filter(hasDefaultAnswer)
-                    .each(setDefaultAnswer).filter(isConcept).each(setDefaultConcept).value();
+                    .each(setDefaultAnswer).value();
+
+                _.chain(defaultsWithAnswers).filter(isConcept).each(setDefaultConcept).value();
+                _.chain(defaultsWithAnswers).filter(isDateTypeAndDefaultValueToday).each(setDefaultTodayDate).value();
```
sravya-kanagarla: @sowmika148 Filtering with isDateType first and then setting value will divide the responsibilities of methods.

PR #254, opened 2019-09-15
BAH-663 | Sowmika | Weekly View in appointments calendar

sowmika148:


ui/app/appointments/constants.js
```diff
@@ -24,13 +25,17 @@ Bahmni.Appointments.Constants = (function () {
         defaultCalendarEndTime: "19:00",
         defaultExpandServiceFilter: -1,
         collapseServiceFilter: 0,
+        defaultWeekDays: [0, 1, 2, 3, 4, 5, 6],
```
angshu: What does 0,1,2 .. mean - for defaultWeekDays and defaultWeekStartDay? if its supposed to be Sunday, Monday ...? I also see that below you have defaultWeekStartDayName as the day name. 
Why separate "defaultWeekStartDay" and "defaultWeekStartDayName"? 
Wouldn't it better, to create objects like :

const weekdays = {
    SUNDAY: 'SUNDAY',
    MONDAY: 'MONDAY',
    ....
}

like below. 

Seems to be redundant definition/usage of the same thing.

ui/app/appointments/constants.js
```diff
@@ -24,13 +25,17 @@ Bahmni.Appointments.Constants = (function () {
...
+        defaultWeekStartDay: 0,
+        defaultWeekStartDayName: 'Sunday',
         minDurationForAppointment: 30,
         appointmentStatusList: [ "Scheduled", "CheckedIn", "Completed", "Cancelled", "Missed" ],
         regexForTime: /^(?:(?:1[0-2]|0?[1-9]):[0-5]\d\s*[AaPp][Mm])?$/,
         privilegeManageAppointments: 'app:appointments:manageAppointmentsTab',
         privilegeForAdmin: 'app:appointments:adminTab',
         privilegeOwnAppointments: 'Manage Own Appointments',
         availableForAppointments: 'Available for appointments',
+        weekDays: {"Monday": 1, "Tuesday": 2, "Wednesday": 3, "Thursday": 4, "Friday": 5, "Saturday": 6, "Sunday": 7 },
```
angshu: See above for related comment. If you use an objects for constants like this, you should leverage the above definition. Also, in this line - there is no "0", however for above "defaultWeekDays", it starts with "0".

ui/app/appointments/constants.js
sowmika148: This is angular calendar days of week. Changed the variable name to angularCalendarDaysOfWeek. We have to mention all the working days here. So it would be 0-6 all numbers.

ui/app/appointments/constants.js
sowmika148: Used only one variable for getting the default value

bsneha90:
Resolving all conflicts, the requested review comments have been incorporated.

PR #260, opened 2019-11-15
BAH-790 Functionality to view obs forms based on provider Types

krishnanspace:
To test this functionality this PR is also needed: https://github.com/Bahmni/default-config/pull/38
This functionality depends on 2 configs mentioned in the above PR.

The first one is `enableProviderTypeBasedFormsAccess` which is set to false by default. When its set to false, the regular functionality/workflow takes place i.e. All the Obs Forms are visible to the current provider.

If it is set to true , the Provider Type Based access to forms is enabled i.e. a particular provider is only able to view the Obs forms assigned to them.

In the linked PR , there are 2 provider types: provider1 & provider2 and each have a particular set of forms that they can view.

For this to actually work, providerAttribute types have to be created in OpenMRS with the exact same names as the ones specified in the app.json file. The attribute datatype must be boolean.

Create multiple attributes as per your requirements. Then set the value of the attributes to true for your providers. For example, your Provider is called "Dr.John" who is of type "XYZ" then set the XYZ provider attribute for Dr.John to true.

This will ensure that Dr. John of type XYZ can only view the forms mentioned in the app.json config

### DIFF
--- a/ui/app/clinical/consultation/controllers/conceptSetPageController.js
+++ b/ui/app/clinical/consultation/controllers/conceptSetPageController.js
@@ -10,11 +10,16 @@ angular.module('bahmni.clinical')
             $scope.consultation.selectedObsTemplate = $scope.consultation.selectedObsTemplate || [];
             $scope.allTemplates = $scope.allTemplates || [];
             $scope.scrollingEnabled = false;
+            var providerTypeBasedFormsAccess;
             var extensions = clinicalAppConfigService.getAllConceptSetExtensions($stateParams.conceptSetGroupName);
             var configs = clinicalAppConfigService.getAllConceptsConfig();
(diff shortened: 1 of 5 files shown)


ui/app/common/domain/services/privilegeTypeService.js
```diff
@@ -0,0 +1,26 @@
...
+
+        var getCurrentUserPrivilegeName = function (privilege, currentUserPrivileges) {
+            var privilegeNames = [];
+            _.filter(_.map(currentUserPrivileges, function (currentObj) {
+                if (currentObj.name.includes(privilege)) {
+                    var match = currentObj.name.split(":viewFormsFor");
+                    privilegeNames.push(match[1]);
+                }
+            }));
+            return privilegeNames;
```
binduak: @krishnanspace What exactly does **privilegeNames** return ??

PR #273, opened 2020-01-20
#BAH-973 | Vinay | Print the List View in Bed Management module

vinay-ThoughtWorks:


ui/app/bedmanagement/views/wardListPrint.html
```diff
@@ -0,0 +1,109 @@
+<style>
+    @page {
+        size: A4 landscape;
+    }
+
+    body {
+        font-family: sans-serif;
+        font-size: 10px !important;
```
bsneha90: Can we avoid using important?

vinay-ThoughtWorks: Done

bsneha90: You have removed 'important' without any other change, does it still work?

vinay-ThoughtWorks: Yes. I checked. I found no change

PR #281, opened 2020-03-08
Fixing verison dependency during "npm install" command.

AdamSoddo:
This change fixes the error:
ReferenceError: primordials is not defined
...
npm ERR! Failed at the jpegtran-bin@0.1.7 postinstall script.
by changing the version of "grunt-contrib-imagemin" to "~2.0.0"

This fix makes it possible to follow the instructions on:
    https://bahmni.atlassian.net/wiki/spaces/BAH/pages/32604585/Setting+up+Bahmni+Dev+Environment


ui/package.json
```diff
@@ -34,6 +34,7 @@
     "grunt-regarde": "~0.1.1",
     "grunt-rename": "~0.1.3",
     "grunt-usemin": "~2.2.0",
+    "jpegtran-bin": "^4.0.0",
```
rbuisson: Is this required? I didn't need to specify this dependency for the build to go through.

AdamSoddo: Good catch, that was part of my initial debugging, The error message was regarding jpegtran-bin, so I tried this first before realizing it was a nested dependency that was causing the build error. That line can be removed.

PR #283, opened 2020-03-19
BAH-1009 | Alekhya, Yamuna | Execute the onFormSave event when form save is clicked in consultation

AlekhyaYalla:
Jira link: https://bahmni.atlassian.net/browse/BAH-1009
this pr has dependency on [form controls pr](https://github.com/Bahmni/form-controls/pull/22), please merge this after that pr.

### DIFF
--- a/ui/app/clinical/consultation/controllers/consultationController.js
+++ b/ui/app/clinical/consultation/controllers/consultationController.js
@@ -317,8 +317,9 @@ angular.module('bahmni.clinical').controller('ConsultationController',
             var preSaveEvents = function () {
                 var observationForms = $scope.consultation.observationForms;
                 _.each(observationForms, function (form) {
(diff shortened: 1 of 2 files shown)


ui/app/clinical/consultation/controllers/consultationController.js
```diff
@@ -314,9 +314,17 @@ angular.module('bahmni.clinical').controller('ConsultationController',
                 return getUrl(board);
             };
 
+            var preSaveEvents = function () {
+                var observationForms = $scope.consultation.observationForms;
```
bsneha90, line 318: How are we maikng sure, this runs only with form2?

AlekhyaYalla: in the loop, line 323, we are having a if check  with condition, where only form2 forms will be passed. So it will not script will not be run for form1

PR #287, opened 2020-04-02
Sneha Bagri | BAH-754 | Upgrade the react version to 16

bsneha90:
This depends on. - https://github.com/Bahmni/form-controls/pull/21. The build will fail until this is merged and published.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -41,9 +41,7 @@
     "karma-jasmine": "~0.2.0",
     "karma-junit-reporter": "^0.2.2",
     "karma-ng-html2js-preprocessor": "^0.1.2",
-    "matchdep": "^0.3.0",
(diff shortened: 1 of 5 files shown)


ui/package.json
```diff
@@ -65,5 +63,9 @@
   "readmeFilename": "README.md",
   "bugs": {
     "url": "https://github.com/Bhamni/opd/issues"
+  },
+  "dependencies": {
+    "react": "^16.12.0",
```
sumanmaity112, line 68: @bsneha90, is it possible to use [latest react version](https://reactjs.org/versions/)

bsneha90: React issues a new release every week, when this was done we did take the latest. As a product it will be difficult to keep up to speed with that.

PR #344, opened 2021-03-12
BAH-1164 | Redirect to login page upon 403

rbuisson:
JIRA ticket: https://bahmni.atlassian.net/browse/BAH-1164
Talk thread: https://talk.openmrs.org/t/bahmni-error-403-handling/32509?u=mksrom


rbuisson:
This introduces some serious bugs. We will revisit the fix.

PR #346, opened 2021-03-30
Bah 1161 - Internationalization UAT bug fixes and label changes

buvaneswari-arun:


ui/app/adt/controllers/adtController.js
```diff
@@ -29,7 +29,23 @@ angular.module('bahmni.adt')
...
 
+                    var keyName = attribute.toUpperCase().replace(/\s\s+/g, ' ').replace(/[^a-zA-Z0-9 _]/g, "").trim().replace(/ /g, "_");
+                    var translationKey = keyPrefix + "_" + keyName;
+                    var translation = $translate.instant(translationKey);
+                    if (translation != translationKey) {
+                        attribute = translation;
+                    }
+                }
+                return attribute;
+            };
```
binduak, line 48: @buvaneswari-arun we already have this method in the translationUtil.js. Then why are we not using it here ?

buvaneswari-arun: TranslationUtils.translateAttribute method is not accessible from this bedmanagement. Hence forced to put the content of the method directly here.

binduak: @buvaneswari-arun is it the same case with other modules ?

buvaneswari-arun: I tried with Bedmanagement - as it was part of the. card - not sure about IPD and programs etc and they will be played in the upcoming releases

binduak (approved):

PR #357, opened 2021-05-11
BAH-1201 | Checking whitelisted domains if configured, otherwise checking for current window origin before redirection

angshu:


ui/app/home/controllers/loginController.js
```diff
@@ -186,13 +186,30 @@ angular.module('bahmni.home')
                     });
                 };
 
+                var urlMatches = function (domain, url) {
+                    if (!domain.trim().endsWith("/")) {
```
buvaneswari-arun, line 190: before domain..trim() Should we be checking for null ?

PR #376, opened 2021-09-14
Tarun | MOBN-1739 | Fix UI alignment in consultation form for tabular section

tarunkumar-tw:
**Jira Link:** https://bahmni.atlassian.net/browse/BAH-1245

### DIFF
--- a/ui/app/styles/clinical/_observations.scss
+++ b/ui/app/styles/clinical/_observations.scss
@@ -428,6 +428,11 @@ concept-set {
     padding-bottom: 30px;
     display: inline-block;
   }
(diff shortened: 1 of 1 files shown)


ui/app/styles/clinical/_observations.scss
```diff
@@ -428,6 +428,11 @@ concept-set {
     padding-bottom: 30px;
     display: inline-block;
   }
+  table {
```
angshu, line 431: this will apply to all table elements, wouldn't it? why not use a class and apply it at appropriate place?

PR #377, opened 2021-09-17
BAH-1240 | Removing check for response head. Oviyam sends a redirect URL which can be HTTP depending on setup, and in that case browser thrrows error for CORS

angshu:


gsluthra:
[From Angshu] We were checking using ajax request for HEAD to see if there was any result. However, oviyam being behing apache and sees the connection coming in in plain HTTP, and the redirection (302) can not be followed by the browser because of same origin policy (which we put in this release). As a result the study is not opened up.

gsluthra (approved):

PR #383, opened 2021-11-09
BAH-1161 | Fixing issue of only drug concept name being shown. Also guarding against empty drug name

angshu:


gsluthra (approved):

gsluthra:
Looks good to me Angshu. thanks!

PR #405, opened 2022-03-31
Refactor. ExtraIdentifier translation in edit patient 

SanoferSameera:


ui/app/registration/views/patientcommon.html
```diff
@@ -157,7 +157,7 @@
                 <div ng-if="patient.uuid != undefined">
                     <article ng-repeat="identifier in patient.extraIdentifiers" class="form-field existing-patient">
                         <div class="field-attribute">
-                            <label class="control-label" for="patientID">{{:: getTranslatedPatientIdentifier(identifier.identifierType.name)}}
```
binduak, line 160: @SanoferSameera what is the diff between getTranslatedPatientIdentifier() and getTranslatedPatientControls() methods ?

binduak:
@SanoferSameera PR has commit details which are not relevant to this. Can you make sure we have only required commits as part of the PR.

ui/app/registration/views/patientcommon.html
swatigogia2020: @binduak both the methods are doing executing exactly the same code, but getTranslatedPatientIdentifier is defined in createPatientController.js (hence is inaccessible for edit patient), while getTranslatedPatientControls is define in patientCommonController.js( hence is accessible for edit patient)

PR #410, opened 2022-05-05
BAH-1546 | Helm charts for Bahmni

stavtnameh:
Created and tested helm charts for Bahmni-web.


stavtnameh:
changed chart name to bahmni-web.

PR #416, opened 2022-05-27
BAH-1816 | Disabling search engine crawling/indexing

umair-fayaz:
Co-authored-by: stavtnameh <hemanth.v@beehyv.com>

### DIFF
--- a/package/docker/httpd.conf
+++ b/package/docker/httpd.conf
@@ -29,7 +29,7 @@
 # least PidFile.
 #
 ServerRoot "/usr/local/apache2"
-X-Robots-Tag: noindex, nofollow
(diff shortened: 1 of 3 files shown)


package/docker/httpd.conf
```diff
@@ -29,7 +29,7 @@
 # least PidFile.
 #
 ServerRoot "/usr/local/apache2"
-
+X-Robots-Tag: noindex, nofollow
```
mohan-13, line 32: I think we would need to add as a header attribute. https://developers.google.com/search/docs/advanced/robots/robots_meta_tag#xrobotstag-implementation

mohan-13 (approved):

PR #421, opened 2022-07-07
BAH-1973 | Add Navigation link for Lab Lite

abinaya-u:


ui/app/i18n/clinical/locale_en.json
```diff
@@ -196,6 +196,7 @@
     "POP_UP_CLOSE_DIALOG_MESSAGE_KEY":"You might lose unsaved data. Are you sure you want to leave this page?",
     "MESSAGE_DIALOG_OPTION_COPY" : "Copy Error",
     "MESSAGE_DIALOG_OPTION_OKAY": "OK",
+    "LAB_LITE_KEY" : "Lab Lite",
```
N0-man, line 199: If we are calling it `Lab Entry` on the Clinical services tile then we should use the same terminology here too...

abinaya-u: updated..

PR #431, opened 2022-08-14
BAH-1720|configure-the-patient-queues-active-and-all-to-search-the-patients

shilpa-iplit:
BAH-1720|configure-the-patient-queues-active-and-all-to-search-the-patients by ABHA id

### DIFF
--- a/ui/app/common/patient-search/models/search.js
+++ b/ui/app/common/patient-search/models/search.js
@@ -95,12 +95,17 @@ Bahmni.Common.PatientSearch.Search = function (searchTypes) {
             return patient[column];
         }).join(" - ");
 
-        var abhaIdentifier = null;
+        var extraIdentifier = null;
         if (patient.extraIdentifiers) {
-            var objIdentifier = JSON.parse(patient.extraIdentifiers);
-            abhaIdentifier = objIdentifier && objIdentifier["ABHA Address"];
+            var objIdentifiers = JSON.parse(patient.extraIdentifiers);
+            for (var key in objIdentifiers) {
+                extraIdentifier = objIdentifiers[key];
+                break;
+            }
+        } else if (patient.extraIdentifierVal) {
+        	extraIdentifier = patient.extraIdentifierVal;
         }
-        patient.abhaIdentifier = patient.abhaIdentifier ? patient.abhaIdentifier : (abhaIdentifier ? abhaIdentifier : patient.identifier);
+        patient.extraIdentifier = patient.extraIdentifier ? patient.extraIdentifier : (extraIdentifier ? extraIdentifier : patient.identifier);
         patient.image = Bahmni.Common.Constants.patientImageUrlByPatientUuid + patient.uuid;
         return patient;
     }
--- a/ui/app/common/patient-search/models/search.js
+++ b/ui/app/common/patient-search/models/search.js
@@ -95,6 +95,12 @@ Bahmni.Common.PatientSearch.Search = function (searchTypes) {
             return patient[column];
         }).join(" - ");
 
+        var abhaIdentifier = null;
+        if (patient.extraIdentifiers) {
+            var objIdentifier = JSON.parse(patient.extraIdentifiers);
+            abhaIdentifier = objIdentifier && objIdentifier["ABHA Address"];
+        }
+        patient.abhaIdentifier = patient.abhaIdentifier ? patient.abhaIdentifier : (abhaIdentifier ? abhaIdentifier : patient.identifier);
         patient.image = Bahmni.Common.Constants.patientImageUrlByPatientUuid + patient.uuid;
         return patient;
     }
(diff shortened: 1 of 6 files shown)


gsluthra:
Can we write this code to avoid mentioning "ABHA" in the codebase? This is global code, not india specific. 
Might be nicer if the code is written to use a specific secondary Identifier, so that different people can choose their own identifier (National Id, Social Security, ABHA, etc). 

If we hardcode ABHA in Bahmni codebase, then we are introducing India specific vocabulary in global product. 

Will let @angshu comment on options.

ui/app/common/patient-search/models/search.js
```diff
@@ -95,6 +95,12 @@ Bahmni.Common.PatientSearch.Search = function (searchTypes) {
             return patient[column];
         }).join(" - ");
 
+        var abhaIdentifier = null;
+        if (patient.extraIdentifiers) {
+            var objIdentifier = JSON.parse(patient.extraIdentifiers);
+            abhaIdentifier = objIdentifier && objIdentifier["ABHA Address"];
```
gsluthra, line 101: @angshu - needed some suggestion from you. Can we instead of hardcoding "ABHA Address" in code, make this a configuration field, and then use that `extraIdentifier` to search. So the configuration field could be something like: "searchByExtraIdentfier=<attribute_name>", and then in Clinic_Config india deployment they can set this to `ABHA Address`. This way the code will not mentioned ABHA anywhere (since this is Bahmni Global code, and not india specific).

ui/app/common/patient-search/models/search.js
```diff
@@ -95,6 +95,12 @@ Bahmni.Common.PatientSearch.Search = function (searchTypes) {
...
+        }
+        patient.abhaIdentifier = patient.abhaIdentifier ? patient.abhaIdentifier : (abhaIdentifier ? abhaIdentifier : patient.identifier);
```
gsluthra, line 103: In this case, the variable would be called: `patient.extraIdentifier`

shilpa-iplit: Can we pick up first extra identifier from objIdentifier object?

gsluthra: We need the name of the Identifier from configuration. Maybe a global property which says.. which extraIdentifier should be searched on? As per my understanding, people may configure multiple identifiers, but want to enable a search only from one of the them (like PAN, or ABHA, or SSN, etc). We should let them choose which one will act as the search identifier (besides the primary one). Sometimes people may want NONE, in which case this configuration will be NULL.

shilpa-iplit: This code is in model file - openmrs-module-bahmniapps/ui/app/common/patient-search/models/search.js
       var extraIdentifier = null;
        if (patient.extraIdentifiers) {
            var objIdentifiers = JSON.parse(patient.extraIdentifiers);
            for (var key in objIdentifiers) {
                extraIdentifier = objIdentifiers[key];
                break;
            }
        } else if (patient.extraIdentifierVal) {
            extraIdentifier = patient.extraIdentifierVal;
        }
        patient.extraIdentifier = patient.extraIdentifier ? patient.extraIdentifier : (extraIdentifier ? extraIdentifier : patient.identifier);
It will pick up the first identifier in the extraIdentifiers. If not present then primary identifier will be displayed.
Please advise.

ui/app/common/patient-search/models/search.js
```diff
@@ -95,6 +95,17 @@ Bahmni.Common.PatientSearch.Search = function (searchTypes) {
...
         }).join(" - ");
 
+        var extraIdentifier = null;
+        if (patient.extraIdentifiers) {
+            var objIdentifiers = JSON.parse(patient.extraIdentifiers);
+            for (var key in objIdentifiers) {
+                extraIdentifier = objIdentifiers[key];
+                break;
+            }
+        } else if (patient.extraIdentifierVal) {
```
gsluthra, line 105: Hi @shilpa-iplit , can you explain what this logic means? I didn't understand.

shilpa-iplit: On Line# 99 In case of Clinical Search 'All' tab - search is returning patient.extraIdentifiers with all extra identifiers from bahmni-core code. So here we are picking up the first extra identifier.
On line# 105 For active patients list we are executing query from GP's  "emrapi.sqlSearch.activePatients" which is returning first extra identifier value in model patient as patient.extraIdentifierVal.
GP "bahmni.extraPatientIdentifierTypes" may have list of extra identifiers Here also we are picking up the first extra identifier.
Please advise if all logic is correct.

gsluthra: Will request @binduak or @swatigogia2020 to recheck. I think its right.

swatigogia2020 (approved):

gsluthra (approved):

PR #440, opened 2022-09-06
BAH-2065 | Hiding the unsupported browser message for Firefox

kushboojain:


ui/app/home/controllers/loginController.js
```diff
@@ -71,8 +71,9 @@ angular.module('bahmni.home')
                 });
             });
 
-            $scope.isChrome = function () {
-                if ($window.navigator.userAgent.indexOf("Chrome") != -1) {
+            $scope.isChromeOrFirefox = function () {
```
N0-man, line 74: would be better to keep the function names agnostic of the impl... so `isSupportedBrowser`

gsluthra:
An interesting read: https://code-boxx.com/detect-browser-with-javascript/

For now, I think checking the userAgent is fine for us. thanks!

gsluthra (approved):

PR #442, opened 2022-09-14
BAH-2283 | Refactor build workflow to use versioning utility

mohan-13:
- Uses `setArtifactVersion` utility from bahmni-infra-utils
- Merge helm package publish job into a single job


mohan-13:
Changes incorporated along with #446

PR #457, opened 2022-09-26
BAH-2306 | Show different range for lab-test

swe-vmd:


swe-vmd:
@binduak is this okay ? if yes we will close the old one.

PR #459, opened 2022-09-29
BAH-2292 | Removed default-config from docker image

mohan-13:
- Removes packaging of default-config in docker image
- Added volumeMount for Helm configuration
- Removed download_artifact.sh as no artifact is consumed from Github Actions


mohan-13:
PR for docker-compose https://github.com/Bahmni/bahmni-package/pull/133

PR #473, opened 2022-10-25
Siva Reddy | BAH-2464 | Refactoring patient search code to bahmni-commons

sivareddytw:
JIRA - [https://bahmni.atlassian.net/browse/BAH-2464](https://bahmni.atlassian.net/browse/BAH-2464)

- Updating Patient Search End Point


ui/app/common/constants.js
```diff
@@ -9,6 +9,7 @@ Bahmni.Common = Bahmni.Common || {};
     var RESTWS = hostUrl + "/openmrs/ws/rest";
     var RESTWS_V1 = hostUrl + "/openmrs/ws/rest/v1";
     var BAHMNI_CORE = RESTWS_V1 + "/bahmnicore";
+    var BAHMNI_COMMONS = RESTWS_V1 + "/bahmni";
```
binduak, line 12: @sivareddytw the key doesn't match with value. is the value should be "bahmnicommons" ?

sivareddytw: @binduak Angshu suggested to use the end point with `/bahmni` instead of `/bahmnicommons`.

binduak (approved):

PR #489, opened 2022-12-05
Tarun,Supriya,Ramu, Prajakta, Venky | BAH-1389 | MOBN-1748 | MOBN-1784 | Registration Module] Make local language field mandatory and add UI Validation to extra Patient Identifiers

venkybavisetti:


ui/app/registration/views/patientcommon.html
```diff
@@ -152,92 +152,7 @@
 
         <section>
             <!--edit patient flow-->
-            <section>
-                <div ng-if="patient.uuid != undefined">
```
angshu, line 156: Why are these code sections removed?

ui/app/registration/controllers/patientCommonController.js
```diff
@@ -223,6 +223,16 @@ angular.module('bahmni.registration')
                 }
             });
 
+            $scope.localLanguageNameIsRequired = function (nameType) {
```
angshu, line 226: I am not sure if this is advisable anymore i.e using person attribute. OMRS already has a name model where names in different locale can be added. I would have advised that. 

Also, just because the person attribute is marked as required does not mean that value must be provided - I would have thought you enable this by some sort of configuration (which I think we already have).

venkybavisetti: As discussed, we couldn’t see any option make the field has mandatory in the back-end so we have added

angshu (approved):

binduak:
@venkybavisetti waiting for bahmni-lite beta release. once its done this week I will go ahead with merge

binduak (approved):

PR #511, opened 2023-01-23
BAH-2792 | Removed Offset value while saving as this is causing misma…

deepthi-mantena:
…tch while displaying date on UI for type obstoobs flowsheet.


gsluthra:
If this fixes while "saving".. then all the data which has already been saved in current Bahmni implementations (like JSS), won't that still cause issues in obs-to-obs flowsheet?

deepthi-mantena:
Yes @Gurpreet this would not fix the date of existing records.Discussed with Bindu regarding the same will make further changes to address this.

PR #513, opened 2023-01-25
BAH-2763 | added sending prescription in sms feature from dashboard

kavitha-sundararajan:


ui/app/clinical/common/services/treatmentService.js
```diff
@@ -135,13 +135,34 @@ angular.module('bahmni.clinical')
             return deferred.promise;
         };
 
+        var getMessageForPrescription = function (patientData, drugOrderSection) {
```
gsluthra, line 138: We should use a translation string template, not use concatenation to construct the message.

gsluthra: Also, this code I think can be its own javascript class, with a getPrescriptionAsString( ) method, and therefore can also have its own unit test cases, to validate the conversions.

gsluthra: Each drug order can have a getDrugOrderAsString( ) and return the string representation. isn't it?

ui/app/common/domain/services/configurationService.js
```diff
@@ -183,6 +183,30 @@ angular.module('bahmni.common.domain')
...
+                transformResponse: [function (data) {
+                    return data;
+                }]
+            });
+        };
+
+        configurationFunctions.prescriptionEmailToggle = function () {
+            return $http.get(Bahmni.Common.Constants.globalPropertyUrl, {
+                params: {
+                    property: 'sms.enablePrescriptionEmailAlert'
```
gsluthra, line 201: if this is an email field, why does it have the word `sms` in it?

gsluthra: Does it enable some alert message, or is it for deciding if email share option will be shown/not. If so, then it should be called: `bahmni.enableEmailPrescriptionOption`

PR #514, opened 2023-01-25
BAH-2792-updated | Modified to retrieve obsdatetime sent from the API…

deepthi-mantena:
… and display on the UI.

### DIFF
--- a/ui/app/clinical/consultation/mappers/specimenMapper.js
+++ b/ui/app/clinical/consultation/mappers/specimenMapper.js
@@ -21,7 +21,7 @@ Bahmni.Clinical.SpecimenMapper = function () {
 
     this.mapSpecimenToObservation = function (specimen) {
         var observation = {};
-        observation.dateCollected = moment(specimen.dateCollected).format("YYYY-MM-DDTHH:mm:ss.SSSZ");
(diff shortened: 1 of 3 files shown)


ui/app/clinical/consultation/mappers/specimenMapper.js
```diff
@@ -21,7 +21,7 @@ Bahmni.Clinical.SpecimenMapper = function () {
 
     this.mapSpecimenToObservation = function (specimen) {
         var observation = {};
-        observation.dateCollected = moment(specimen.dateCollected).format("YYYY-MM-DDTHH:mm:ss.SSSZ");
+        observation.dateCollected = moment(specimen.dateCollected).format("YYYY-MM-DD");
```
angshu, line 24: any specific reason we are getting rid of the timestamp? time is also very relevant

deepthi-mantena: Functionally, this field is meant to capture the sample collection date. It is captured from a date field(calendar) from the UI so timestamp won't be relevant. Earlier this timestamp was causing timezone offset overhead while saving and retrieving it.

angshu (approved):

PR #517, opened 2023-02-01
BAH-2359 | Showing error pop-up on navigating out of patient dashboard if there are dirty changes 

kushboojain:
This PR also handles  Bah 2771 | Fixing injection issue


ui/app/clinical/consultation/directives/consultationConfirmOnExit.js
```diff
@@ -0,0 +1,50 @@
...
+                    };
+                    $scope.$on('$stateChangeStart', function (event, next, current) {
+                        var noOfMedications = 0;
+                        var navigating = next.url.split("/")[1];
+                        var allConsultationBoards = clinicalAppConfigService.getAllConsultationBoards();
+                        var outOfConsultationBoard = true;
+                        if ($scope.consultation) {
+                            if ($scope.consultation.orders) {
+                                if ($scope.consultation.orders.length !== $scope.consultation.investigations.length) {
+                                    $state.dirtyConsultationForm = true;
```
kk-tw, line 19: This can be rewritten as--
`$state.dirtyConsultationForm = $scope.consultation.orders.length !== $scope.consultation.investigations.length`

ui/app/clinical/consultation/directives/consultationConfirmOnExit.js
```diff
@@ -0,0 +1,50 @@
...
+                                }
+                            }
+                            if ($scope.consultation.newlyAddedTabTreatments && $scope.tabConfigName) {
+                                noOfMedications = $scope.consultation.newlyAddedTabTreatments[$scope.tabConfigName].treatments.length;
+                            }
+                        }
+
+                        var isConsultationFormDirty = noOfMedications > 0 || ($scope[attrs.name] && $scope[attrs.name].$dirty);
+                        if (isConsultationFormDirty) {
+                            $state.dirtyConsultationForm = true;
```
kk-tw, line 29: `$state.dirtyConsultationForm =  isConsultationFormDirty`

ui/app/clinical/consultation/directives/consultationConfirmOnExit.js
```diff
@@ -0,0 +1,50 @@
...
+                        }
+                        allConsultationBoards.map(function (board) {
```
kk-tw, line 31: Please do not use map to loop through an array, a map is generally used to return a new array using its callback function, You an use a simple forEach or for loop in such cases

ui/app/clinical/consultation/directives/consultationConfirmOnExit.js
kushboojain: Yes need to assign it true only if the condition is matched. If the previous value is already true, I don't that to get changed, as it might refer to a dirty check of some other tab.

ui/app/clinical/consultation/directives/consultationConfirmOnExit.js
kushboojain: Updated

kk-tw (approved):

PR #522, opened 2023-02-06
BAH-2763 | sharing prescriptions via sms feature

kavitha-sundararajan:


ui/app/clinical/common/services/treatmentService.js
```diff
@@ -125,6 +125,21 @@ angular.module('bahmni.clinical')
             return deferred.promise;
         };
 
+        var sendPrescriptionSMS = function (visitUuid) {
+            var deferred = $q.defer();
+            var locale = $translate.instant(appService.getAppDescriptor().getConfigValue("smsLanguage") || Bahmni.Clinical.Constants.smsLanguage);
```
angshu, line 130: Not sure about using a property in config just to pick up translation key. Why dont we just hardcode the KEY over here? It will be still be picked from relevant message bundles. Also what happens if the key is not defined in the message bundle, pls use a default in the backend

ui/app/clinical/displaycontrols/treatmentData/directives/treatmentTable.js
```diff
@@ -2,7 +2,9 @@
 
 angular.module('bahmni.clinical')
     .directive('treatmentTable', function () {
-        var controller = function ($scope, $rootScope) {
+        var controller = function ($scope, $rootScope, treatmentService, messagingService) {
+            $scope.prescriptionEmailToggle = $rootScope.prescriptionEmailToggle;
+            $scope.prescriptionSMSToggle = $rootScope.prescriptionSMSToggle;
```
angshu, line 7: why not pass it as attribute to the directive? instead of reading from rootScope

ui/app/clinical/common/services/treatmentService.js
kavitha-sundararajan: refactored it by hardcoding the KEY. If this is not defined, english will be picked up as default in the backend

ui/app/clinical/displaycontrols/treatmentData/directives/treatmentTable.js
kavitha-sundararajan: passed as attributes to directive

PR #529, opened 2023-02-13
updated registration message to read location with tag visit location

kavitha-sundararajan:


ui/app/registration/services/patientService.js
```diff
@@ -91,9 +91,9 @@ angular.module('bahmni.registration')
         };
 
         var getRegistrationMessage = function (patientId, name, age, gender) {
-            var clinicName = $rootScope.loggedInLocation.name;
```
binduak, line 94: @kavitha-sundararajan loggedInLocation and VisistLocation might not be the same all the time

kavitha-sundararajan: Yes @binduak which location is tagged with "Visit Location" tag should be displayed and not the one with "Login Location"

binduak (approved):

ui/app/registration/services/patientService.js
kavitha-sundararajan: Regarding changing the name, sometime back, anghsu/gurpreet suggested that in bahmni-apps we don't need to mention clinic anywhere, so changed the name. but still the message coming from clinic-config has #clinicName with it, so that doesn't make sense to change the name. I've reverted the variable name change it back

PR #531, opened 2023-02-15
Updated to display the name of the doctor who added the observation first 

deepthi-mantena:


ui/app/clinical/common/controllers/visitController.js
```diff
@@ -16,6 +16,24 @@ angular.module('bahmni.clinical')
             $scope.patientUuid = $stateParams.patientUuid;
             $scope.visitUuid = $stateParams.visitUuid;
             var tab = $stateParams.tab;
+            $scope.$on('observations.providers', function (event, observations) {
```
angshu, line 19: do we need to use the $on?

deepthi-mantena: Yes Angshu this will update the provider when there is another visit in same encounter and that visit is started by different doctor in that case UI should be updated with provider who started the visit.So this would capture whenever there is change in provide
Screenshot 2023-02-17 at 6 50 11 PM
Screenshot 2023-02-17 at 6 50 29 PM
Screenshot 2023-02-17 at 6 50 40 PM

angshu (approved):

PR #533, opened 2023-02-16
BAH-2863 | add facility visit location

kavitha-sundararajan:


ui/app/common/domain/services/locationService.js
```diff
@@ -28,10 +28,19 @@ angular.module('bahmni.common.domain')
             });
         };
 
+        var getFacilityVisitLocation = function (locationUuid) {
```
binduak, line 31: @kavitha-sundararajan can you please add tests for the new function

kavitha-sundararajan: added tests

PR #542, opened 2023-02-27
Deepthi | BAH-2849-fix | updated to fetch the provider for the encounter from Visit api

deepthi-mantena:


binduak (approved):

ui/app/clinical/common/controllers/visitController.js
```diff
@@ -16,22 +16,18 @@ angular.module('bahmni.clinical')
...
-                        } else if (obs1.observationDateTime > obs2.observationDateTime) {
-                            return 1;
-                        } else {
-                            return 0;
-                        }
+            visitService.getVisit($scope.visitUuid, 'custom:(uuid,visitType,startDatetime,stopDatetime,encounters:(uuid,encounterDatetime,provider:(display),encounterType:(display)))').then(function (response) {
+                if (response.data && response.data.encounters) {
+                    var encounters = response.data.encounters;
+                    encounters = encounters.filter(function (enc) {
+                        return enc.encounterType.display == 'Consultation';
```
angshu, line 24: Any specific reason we are hardcoding the "encounter type"? what happens if people want to include other types of encounters? I think this should be configure in app config .. and also implementation should be able to specify a list (array) of encounter types.

deepthi-mantena: @Angshu have made the changes from clinic config and updated the above code accordingly.Please verify

ui/app/clinical/common/controllers/visitController.js
```diff
@@ -16,22 +16,19 @@ angular.module('bahmni.clinical')
             $scope.patientUuid = $stateParams.patientUuid;
             $scope.visitUuid = $stateParams.visitUuid;
             var tab = $stateParams.tab;
-            $scope.$on('observations.providers', function (event, observations) {
-                var consultationObservations = observations.filter(function (obs) {
-                    return obs.encounterTypeName === 'Consultation';
-                });
+            var encounterType = appService.getAppDescriptor().getConfigValue("encounterAttributes").encounterType;
```
angshu, line 19: What would a sample config look like? 
- name should be well understood - maybe - 'filterEncounterTypes'?
- should be an array and is optional, maybe non-existent. 
- the app can by default scan all encounter type or 'consultation' encounterType if not configured in app-config

angshu (approved):

PR #544, opened 2023-03-01
BAH-2189 | fix for email global property, blank page, email subject

kavitha-sundararajan:


ui/app/common/domain/services/configurationService.js
```diff
@@ -195,6 +195,18 @@ angular.module('bahmni.common.domain')
             });
         };
 
+        configurationFunctions.prescriptionEmailToggle = function () {
```
binduak, line 198: @kavitha-sundararajan what does prescriptionEmailToggle holds and how are we using this to drive the functionality here ?

kavitha-sundararajan: @binduak prescriptionEmailToggle is a global property for on and off of email functionality. Only if it is set as true, the email icon will be visible in the share options UI. By default it will be false and the icon won't be visible under share options.

binduak: why it can't be a configuration ? we are validating this value on both UI/backend ?only on UI ?

kavitha-sundararajan: We are validation this value on UI only. Long back, Gurpreet suggested this to be a global property instead of configuration because if we kept it as config, then it would require re-deployment of clinic-config and this should be changeable in the UI itself.

binduak: no @kavitha-sundararajan i am still not convinced as this property is not being used in the backend. And this is still can be a config parameter.

angshu: in whatever case, backend must check whether the property is enabled to send email and do so accordingly.

binduak: @kavitha-sundararajan yes..if we are checking this on backend then it makes sense having it as a global property

angshu: @kavitha-sundararajan are you checking against that that on the backend anyway?

kavitha-sundararajan: @binduak @angshu the validation is not there previously. have updated bahmni-core with the property check before sending email and raised [PR](https://github.com/Bahmni/bahmni-core/pull/194)

binduak (approved):

PR #550, opened 2023-03-14
Parvathy, Rahul | BAH-2887 | Fix. Observation Mapper Node Creation Logic

rahu1ramesh:
JIRA card -> https://bahmni.atlassian.net/browse/BAH-2887
In this PR, chief complaint text format displayed on the consultation tab has been fixed. The chief complaint only displayed the duration. This has been fixed to include the complaint, duration and duration unit in the below format
- "Chief Complaint" + since + "Duration" + "Duration Unit".

Screenshot 2023-04-19 at 7 14 04 PM


ui/app/common/concept-set/mappers/observationMapper.js
```diff
@@ -181,28 +196,75 @@ Bahmni.ConceptSet.ObservationMapper = function () {
...
-
     var getDurationDisplayValue = function (duration) {
-        var durationForDisplay = Bahmni.Common.Util.DateUtil.convertToUnits(duration.value);
+        var durationForDisplay = Bahmni.Common.Util.DateUtil.convertToUnits(
+            duration.value
+        );
         if (durationForDisplay["value"] && durationForDisplay["unitName"]) {
-            return "since " + durationForDisplay["value"] + " " + durationForDisplay["unitName"];
+            return (
+            "since " +
```
gsluthra, line 252: This seems to be ENGLISH text hardcoded. We likely need a translation string with substitution params, to ensure this shows up correctly in other languages. Plz check. thanks!

gsluthra:
Can you attach sample screenshots of Bahmni, to see how it looks on screen now? thanks!

ui/app/common/concept-set/mappers/observationMapper.js
rahu1ramesh: Sure @gsluthra, we'll make the necessary changes.

rahu1ramesh:
> Can you attach sample screenshots of Bahmni, to see how it looks on screen now? thanks!

Screenshot 2023-04-20 at 10 49 44 AM

ui/app/common/concept-set/mappers/observationMapper.js
```diff
@@ -4,26 +4,34 @@ Bahmni.ConceptSet.ObservationMapper = function () {
...
+      $translate
+    ) {
         var observationsForDisplay = [];
         _.forEach(observations, function (savedObs) {
             if (savedObs.concept.conceptClass && (savedObs.concept.conceptClass === Bahmni.Common.Constants.conceptDetailsClassName || savedObs.concept.conceptClass.name === Bahmni.Common.Constants.conceptDetailsClassName)) {
-                var observationNode = new Bahmni.ConceptSet.ObservationNode(savedObs, savedObs, [], savedObs.concept);
-                var obsToDisplay = createObservationForDisplay(observationNode, observationNode.primaryObs.concept);
-                if (obsToDisplay) {
-                    observationsForDisplay.push(obsToDisplay);
+                if (isBahmniLite(observations, $translate)) {
```
gsluthra, line 18: I am not sure why this condition exists? `isBahmniLite()`.

Ideally there should not be any code in Bahmni, that checks for whether Bahmni is running in LITE / STANDARD version. There should be only one code, which is being used in Bahmni UI, and configuration decides what gets shown on screen. 

Please clarify, maybe the variable name is misleading. thanks.

gsluthra (changes requested):

ui/app/common/concept-set/mappers/observationMapper.js
rahu1ramesh: https://github.com/Bahmni/openmrs-module-bahmniapps/blob/1331939dd7e486dac6bb4bc897e70b943685b870/ui/app/common/obs/models/observation.js#L65

The reason we went for this approach where we distinguish between LITE / STANDARD version is due to the fact that the response from Bahmni core varies for both the platforms. We checked how this was previously handled and decided to go ahead with the same. Above is an example of the same where in response to the platform  determines what action is to be taken. The variable name as you suggested could have been done better. Apart from the variable name, the approach is similar to what has been already done in `ui/app/common/obs/models/observation.js#L65`.

gsluthra: > the response from Bahmni core varies for both the platforms
Bahmni LITE runs with a different config folder than Bahmni STANDARD. And implementation partners that adopt Bahmni will modify these configurations to suit their needs. So, in either scenario, responses change depending on configuration (and not if they are LITE/Standard). 

Both Bahmni LITE and STANDARD run on the same OpenMRS+Bahmni codebase.

If we are writing this to ensure backward compatible behavior, then maybe we need to decide what is old behaviour and new behavior, and discuss what methodName/variable name will make sense in code so that it is clearer. Ideally, if this is about translation messages -- we need to ensure old config and new config, both show up as same on screen. 

Happy to discuss further. I might be misunderstanding the platform behaviors/responses. thanks.

rahu1ramesh: Sure @gsluthra, we'll try to address this issue as soon as possible. We will connect with you on the same.

gsluthra (approved):
Good with me. Thanks!

PR #555, opened 2023-03-30
BAH-2944 | exclude default sections in visit dashboard based on config

kavitha-sundararajan:


SooryaKumaranC-tw (approved):

ui/app/clinical/common/services/visitTabConfig.js
```diff
@@ -41,6 +41,12 @@ angular.module('bahmni.clinical')
                 });
                 firstTabWithDefaultSection.sections = _.unionWith(_.values(mandatorySections), _.values(firstTabWithDefaultSection.sections), _.isEqual);
 
+                var tabWithDefaultSectionsExcludes = _.find(tabs, function (tab) { return tab.defaultSectionsExcludes; });
+                var excludedDefaultSections = tabWithDefaultSectionsExcludes ? tabWithDefaultSectionsExcludes.defaultSectionsExcludes : [];
```
angshu, line 45: the property "defaultSectionExcludes" can be undefined or set to null, right? can we put a check against that. Or does the underscrore library ensure that?

kavitha-sundararajan: Yes, it'll be handled even it if is undefined or null. It'll display all the default sections if the config is not defined, set to null or empty

angshu (approved):

PR #564, opened 2023-05-05
BAH-2647 | Already active Visits not allowed to create again

atish160384:
- If the visit of same patient at same location already exists then it is not allowed to create visit again.


ui/app/i18n/registration/locale_en.json
```diff
@@ -87,6 +87,7 @@
   "MESSAGE_DIALOG_OPTION_COPY" : "Copy Error",
   "MESSAGE_DIALOG_OPTION_OKAY": "OK",
   "NO_LOCATION_TAGGED_TO_VISIT_LOCATION": "No location tagged to Visit Location Found",
+  "VISIT OF THIS PATIENT AT SAME LOCATION EXISTS": "Vist of this patient at same location already exists",
```
gokultw, line 90: Keys are having spaces! it will be good if spaces are replaced by underscore

ui/app/registration/directives/patientAction.js
```diff
@@ -43,7 +43,7 @@ angular.module('bahmni.registration')
                     }
                 };
 
-                function setForwardActionKey () {
+                function setForwardActionKey() {
```
gokultw, line 46: Missing space before function parentheses.
Github actions failing.

ui/app/registration/directives/patientAction.js
```diff
@@ -138,24 +138,41 @@ angular.module('bahmni.registration')
...
+                                }
+                            }, function () {
+                                $state.go('patient.edit', { patientUuid: $scope.patient.uuid });
+                            }));
                         }
-                    }, function () {
-                        $state.go('patient.edit', {patientUuid: $scope.patient.uuid});
-                    }));
+                        else {
+                            messagingService.showMessage("error", "VISIT_OF_THIS_PATIENT_AT_SAME_LOCATION_EXISTS");
```
gsluthra, line 173: You move this logic to top as a guard clause. 

if(exists) return "error"

And then everything else doesn't need brackets.

gsluthra:
This check has been added in the **UI codebase**, and not in the **backend**. I think this is incorrect. The UI codebase is just ONE browser client. Ideally this is domain/business logic, which should be in backend, so that once a visit is started by ANY client, no one else should be able to start a visit for the same visit location. I think this code should be in bahmni-core and not in bahmniapps. 

Also, need to validate with Angshuman. Locations can have hierarchy. E.g
Apollo Hospital -> OPD Department -> ClinicA, ClinicB, ClinicC
Apollo Hospital -> IPD Department

There is some expectation that one visit can only be started in one sub-tree. But another visit can be started in another sub-tree. For instance the same patient can have parallel visits in OPD and IPD, but only one visit at a time within OPD/IPD.

gokultw:
We have a different card for Backend fix. https://bahmni.atlassian.net/browse/BAH-1784
We had a discussion with Angshu and decided that we will not fix that. form openmrs it is the expected behaviour to have multiple visits in same location.

ui/app/i18n/registration/locale_en.json
```diff
@@ -87,6 +87,7 @@
   "MESSAGE_DIALOG_OPTION_COPY" : "Copy Error",
   "MESSAGE_DIALOG_OPTION_OKAY": "OK",
   "NO_LOCATION_TAGGED_TO_VISIT_LOCATION": "No location tagged to Visit Location Found",
+  "VISIT_OF_THIS_PATIENT_AT_SAME_LOCATION_EXISTS": "Vist of this patient at same location already exists",
```
gsluthra, line 90: An active visit for this patient already exists for this location. The system cannot create a duplicate visit.

ui/app/i18n/registration/locale_es.json
```diff
@@ -88,6 +88,8 @@
     "MESSAGE_DIALOG_OPTION_OKAY": "OK",
     "NO_LOCATION_TAGGED_TO_VISIT_LOCATION": "No hay ubicación puesta para la Ubicacion de la Visita Encontrada",
     "REGISTRATION_ADDTIONAL_IDENTIFIERS": "Identificadores Adicionales",
+    "VISIT_OF_THIS_PATIENT_AT_SAME_LOCATION_EXISTS": "Ya existe la visita de este paciente en la misma ubicación",
```
gsluthra, line 91: ONLY English translations need to be checked in. Not in other languages. Those will come automatically from Transifex. See this: https://bahmni.atlassian.net/wiki/spaces/BAH/pages/41287695/Dev+Notes+for+Translating+Bahmni

ui/app/registration/directives/patientAction.js
```diff
@@ -138,24 +138,38 @@ angular.module('bahmni.registration')
                     return _.isEmpty($rootScope.visitLocation);
                 };
 
+                var visitExistsCheck = function (patientProfileData) {
```
gsluthra, line 141: Shouldn't this be called: checkIfActiveVisitExists?

ui/app/common/models/visitControl.js
```diff
@@ -18,6 +18,10 @@ Bahmni.Common.VisitControl = function (visitTypes, defaultVisitTypeName, encount
         self.selectedVisitType = visitType;
     };
 
+    self.checkIfVisitExists = function (patientUuid, visitLocationUuid) {
```
gsluthra, line 21: Should this also be called checkIfActiveVisitExists()?

ui/app/common/domain/services/visitService.js
```diff
@@ -37,6 +37,19 @@ angular.module('bahmni.common.domain')
             });
         };
 
+        this.checkVisit = function (patientUuid, visitLocationUuid) {
```
gsluthra, line 40: Should this also be called checkIfActiveVisitExists()?

gsluthra (approved):

PR #566, opened 2023-05-12
BAH-2348 | Pulling in translations

SanoferSameera:


.tx/config
```diff
@@ -1,43 +1,43 @@
 [main]
 host = https://www.transifex.com
 
-[bahmni.admin-app]
+[o:openmrs:p:bahmni:r:admin-app]
```
gsluthra, line 4: Is this provided by transifex? What do they indicate?

SanoferSameera: I have tried without any changes in the first place, but was not able to find the resource. came across the documentation https://developers.transifex.com/docs/using-the-client#adding-resources-to-configuration

.tx/config
kiranmairavipati1978, line 1: @SanoferSameera why haven't we added tx configuration for appointments app? I also see that there are a bunch of apps other than this (eg: documents-upload, ipd, ot). Not clear why are we not interested in pulling translations for these apps from transifex.
cc: @gsluthra @binduak

SanoferSameera: For appointments we have a separate repository, https://github.com/Bahmni/openmrs-module-appointments-frontend

SanoferSameera: I cannot find transifex https://app.transifex.com/openmrs/bahmni/language/fr for those for ipd and ot

gsluthra: We should pull strings that are needed for the green Bahmni EMR UI to work/display correct translations. That will be multiple repos, but mostly i think these would be: bahmniapps, default-config, clinic-config, appointments, bahmni-core, and few omods. We can use a different PR or JIRA ticket for OpenELIS, different for DCM4chee or Odoo. The priority right now will be EMR UI related translations. 

The Bahmni LITE release contains EMR UI, OpenMRS, Backend, etc.. + Crater + ABDM. So pulling those will be helpful for Bahmni LITE release.

gsluthra: (Note: ABDM components are different repo, different ownership, so maybe you can skip ABDM for now. ABDM is for India only).

gsluthra: For IPD / OT you can ping @binduak or @buvaneswari-arun to check if they know. Likely these files were never uploaded to transifex. I see ADT, but not OT. Also, IPD/OPD share their files in many cases, because dashboards are re-used, or screens are re-used from CLINICAL.

gsluthra:
I checked translations for various languages by randomly picking and putting them in google translate, and got accurate translations in all cases. Seems good to me for merging. thanks!

PR #567, opened 2023-05-12
BAH-2999 | Add. translation for italian

SanoferSameera:


ui/app/i18n/admin/locale_it.json
kiranmairavipati1978, line 1: @SanoferSameera did you first create empty json files for each of the apps and then ran `tx pull -l it`
or would tx pull automatically create json files(locale-it.json) based o .tx configuration?

SanoferSameera: just pull creates a file as well

gsluthra: Ideally just pull should create the files automatically. Most often was projects do is that they only create ENGLISH files in their codebase, and other files are pulled in via transifex in the build process. But, since we don't do this daily/automatically, we are doing a manual pull of language files and committing them into code.

Also, this is why there is likely no need for people to edit the non-english files in Github, because they will get over-written when new translations are pulled in from transifex. Thats' my understanding.

gsluthra (approved):
I used google translate to verify 5 random translations, and the english translation was accurate for all of them. So, I am assuming this is good to go. Thanks!

PR #568, opened 2023-05-15
BAH-3000| Add. translation for chinese

SanoferSameera:


gsluthra:
I used google translate to verify 5 random translations, and the english translation was accurate for all of them. So, I am assuming this is good to go. Thanks!

gsluthra (approved):

PR #598, opened 2023-06-09
BS-148 | Adding support for multiple prefix search in drug order

sivareddyp:
JIRA => https://bahmni.atlassian.net/browse/BS-148

- Even though, api call `openmrs/ws/rest/v1/drug?q=Diph+hyd&s=ordered&v=..` fetching results, we are still not returning the drugs only prefix matches.

Screenshot 

image

### DIFF
--- a/ui/app/clinical/consultation/models/drugSearchResult.js
+++ b/ui/app/clinical/consultation/models/drugSearchResult.js
@@ -18,7 +18,9 @@ Bahmni.Clinical.DrugSearchResult = (function () {
     var getMatcher = function (searchString) {
         return function (value) {
             // return value.search(new RegExp(searchString, "i")) !== -1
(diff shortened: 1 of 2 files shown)


vijayanandtwks (approved):

gsluthra:
Would this be also a good idea to have in Order Set Create screen, where drugs are searched and added to the order set. See: https://bahmni.atlassian.net/wiki/spaces/BAH/pages/49545249/Authoring+Order+Sets

Code wise seems good to me.

gsluthra (approved):

sivareddyp:
@gsluthra prefix search working in order set page. Here, we are only making API call `openmrs/ws/rest/v1/drug?q=Diph+hyd&s=ordered&v=.` but, in **drug order** page, we are doing processing on the response of API call.

image

PR #617, opened 2023-06-23
Phani, Sowmya | BAH-3078 | Configures custom person attribute and adds kid icon

sowmya-AS:
This PR
1. modifies patient search api to return custom person attribute
2. adds kid icon to patient dashboard , clinical and program queues, ward list  etc alongside patient details based on custom person attribute value returned 

JIRA Reference:[ Add kid icon]( https://bahmni.atlassian.net/jira/software/c/projects/BAH/boards/55?modal=detail&selectedIssue=BAH-3078 )

### DIFF
--- a/ui/app/adt/models/wardDetails.js
+++ b/ui/app/adt/models/wardDetails.js
@@ -4,7 +4,7 @@ Bahmni.ADT.WardDetails = {};
 
 Bahmni.ADT.WardDetails.create = function (details, diagnosisStatus) {
     var detailsMap = {};
-    var attributesToCopy = ["Bed", "Ward", "Id", "Name", "Age", "Gender", "District", "Village", "Admission By", "Admission Time", "Disposition By", "Disposition Time", "ADT Notes"];
+    var attributesToCopy = ["Bed", "Ward", "Id", "Name", "Age", "Gender", "District", "Village", "Admission By", "Admission Time", "Disposition By", "Disposition Time", "ADT Notes", "kid"];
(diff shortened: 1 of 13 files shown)


binduak:
@Phanindra-tw @sowmya-AS we need JIRA card link here in the PR header and description. Please check

ui/app/adt/models/wardDetails.js
```diff
@@ -4,7 +4,7 @@ Bahmni.ADT.WardDetails = {};
 
 Bahmni.ADT.WardDetails.create = function (details, diagnosisStatus) {
     var detailsMap = {};
-    var attributesToCopy = ["Bed", "Ward", "Id", "Name", "Age", "Gender", "District", "Village", "Admission By", "Admission Time", "Disposition By", "Disposition Time", "ADT Notes"];
```
binduak, line 7: @Phanindra-tw can you check changing the relevant sql query would bring the kid value instead of adding in the headers list ?

binduak:
@Phanindra-tw what happens if person attribute name is other than "kid" or if someone doesn't configure the person attribute "kid" ? are we sure that the application doesn't break anywhere, have you checked this scenario?

ui/app/adt/models/wardDetails.js
Phanindra-tw: @binduak the variable "attributesToCopy" will not be used to query from backend , In the backend response if we have kid as attribute it will copy otherwise it will be as undefined

Phanindra-tw: The backend is sending the kid attribute we added the sql statements for the same to return the "kid" attribute, the frontend is only accepting the keys mentioned in `attributesToCopy` array. So we need to add "kid" here.

binduak (approved):

PR #640, opened 2023-07-13
BAH-3056 | Calculate rule based dosage

SooryaKumaranC-tw:
Jira Link - https://bahmni.atlassian.net/browse/BAH-3056


rahu1ramesh:
The prescribed medicine dosage will now be calculated based solely on `weight` and only if the medicine is in `milligrams ("mg")`, as per CURE's requirement. However, this implementation may lead to unexpected results or failure if the rule involves `mg/m2` against `height` or if it's a `custom rule`. Additionally, the same failure would apply to any medication that isn't measured in `milligrams`. 

I had a discussion with @SooryaKumaranC-tw, @binduak, and @maharaja-m earlier today, and we have two possible approaches:

1. **Rewrite the dosage calculation logic to handle all the mentioned scenarios.**
2. **Temporarily support only milligrams per kilogram (mg/kg) and hide the options for mg/m2 and custom rules. This means the dosage calculation will only work for medications with "mg" as their unit. (CURE recommends this approach for now. A screenshot from their environment illustrating this has been attached below for reference).**

We're waiting for your opinion on which approach to take, @angshu. The PR is currently on hold.

cc: @parvathy00, @SanoferSameera, @gokultw

rahu1ramesh (changes requested):

parvathy00 (approved):

parvathy00 (approved):

PR #648, opened 2023-07-21
Sowmya|Phani | BAH-3113 : Adds React components for the right panel of Drug Chart(MAR)

sowmya-AS:
Link to JIRA card -
https://bahmni.atlassian.net/browse/BAH-3113


bassoGeorge (approved):
Looks good

PR #671, opened 2023-08-18
[Sowmya|Tanya] | Adds logic to fetch drug chart data | A-1204950368410820

tanyaa-tw:


Arjun-Go (approved):

micro-frontends/src/ipd/IpdDashboard.jsx
```diff
@@ -19,11 +18,7 @@ export function IpdDashboard(props) {
...
   hostData: PropTypes.shape({
-    patient: PropTypes.shape({
-      uuid: PropTypes.string.isRequired,
-    }).isRequired,
-  }),
-  hostApi: PropTypes.shape({
-    onConfirm: PropTypes.func,
-  }),
+    patientId: PropTypes.string,
+    forDate: PropTypes.instanceOf(Date),
```
Phanindra-tw, line 22: Do we need to send the forDate from Bahmni Apps?
I think the navigation of dates will be driven from IPD module only. So I think sending only patientId would be sufficient.

Arjun-Go:
Since you don't need the confirmBox in the patientDasboardController, the mocks for the confirmbox can be removed from the test file as well. - https://github.com/Bahmni/openmrs-module-bahmniapps/blob/01b30fcbb51e2d39f048473a8be709f7198d1966/ui/test/unit/clinical/controllers/patientDashboardController.spec.js#L46

ui/app/clinical/dashboard/controllers/patientDashboardController.js
```diff
@@ -25,29 +25,10 @@ angular.module('bahmni.clinical')
...
-                        };
-
-                        confirmBox({
-                            scope: dialogScope,
-                            actions: [{ name: "okay", display: "Okay" }],
-                            className: "ngdialog-theme-default"
-                        });
-                    }
+                    patientId: $stateParams.patientUuid,
+                    forDate: new Date().toUTCString()
```
Phanindra-tw, line 31: Same as abve

tanyaa-tw: We are directly passing the epoch value of date from bahmniapps. In IPD-frontend repo we are directly receiving the epoch value for forDate

Phanindra-tw: Okay, for now lets keep it like this, We might have to revisit this when navigation card is played.

Phanindra-tw (approved):

PR #681, opened 2023-08-25
BAH-3179 | Config for displaying patient search results

umair-fayaz:
Documentation: https://bahmni.atlassian.net/wiki/spaces/BAH/pages/3221684257/Patient+Search+by+Phone+Number+in+Registration+Module

[Jira: ](https://bahmni.atlassian.net/browse/BAH-3179)

### DIFF
--- a/ui/app/registration/controllers/searchPatientController.js
+++ b/ui/app/registration/controllers/searchPatientController.js
@@ -5,8 +5,12 @@ angular.module('bahmni.registration')
         'messagingService', '$translate', '$filter',
         function ($rootScope, $scope, $location, $window, spinner, patientService, appService, messagingService, $translate, $filter) {
             $scope.results = [];
+            var searchDisplayOptions = appService.getAppDescriptor().getConfigValue("patientSearchResultOptions") || {};
+            var ignoredIdentifiers = new Set(searchDisplayOptions.ignorePatientIdentifiers || []);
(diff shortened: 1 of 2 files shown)


ui/app/registration/controllers/searchPatientController.js
```diff
@@ -5,8 +5,12 @@ angular.module('bahmni.registration')
         'messagingService', '$translate', '$filter',
         function ($rootScope, $scope, $location, $window, spinner, patientService, appService, messagingService, $translate, $filter) {
             $scope.results = [];
+            var searchDisplayOptions = appService.getAppDescriptor().getConfigValue("patientSearchResultOptions") || {};
+            var ignoredIdentifiers = new Set(searchDisplayOptions.ignorePatientIdentifiers || []);
+            $scope.showAge = searchDisplayOptions.showAge != null ? searchDisplayOptions.showAge : true;
+            $scope.showDOB = searchDisplayOptions.showDOB != null ? searchDisplayOptions.showDOB : false;
```
kavitha-sundararajan, line 11: instead of this way, can we just set it like `$scope.showAge = searchDisplayOptions.showAge` ? can you confirm if that works

umair-fayaz: We also need a default behaviour Kavitha, incase the config is not present. we should then display only age

kavitha-sundararajan (approved):

PR #687, opened 2023-09-01
BS-208 | Placement of Alert near Diagnosis

nyatindopatrick:
Jira: https://bahmni.atlassian.net/browse/BS-208

## Summary
Display already dismissed alert message against the saved/draft Diagnosis 

Screenshot 2023-09-01 at 09 31 40

### DIFF
--- a/ui/app/clinical/common/services/cdssService.js
+++ b/ui/app/clinical/common/services/cdssService.js
@@ -1,0 +1,229 @@
+'use strict';
+
+angular.module('bahmni.clinical')
+    .service('cdssService', ['drugService', function (drugService) {
+        var createMedicationRequest = function (medication, patientUuid, conceptSource) {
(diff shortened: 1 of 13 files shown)


ui/app/clinical/common/services/cdssService.js
```diff
@@ -0,0 +1,229 @@
+'use strict';
+
+angular.module('bahmni.clinical')
+    .service('cdssService', ['drugService', function (drugService) {
```
gsluthra, line 4: In general there is a lot of "code" that is written for CDSS functionality to work. I think maybe we can add a check for whether cdss functionality is enabled / not first.. and if its not enabled, we don't need to register any of the methods/listeners --- to keep page performance / events/etc at a minimal. What do you think? 

//At the start of the code. 
if (cdss == false) return;

gsluthra (approved):

PR #696, opened 2023-09-10
BAH-3049 | Samridhi | Put default dates in Report screen to save time

Samridhi-98:
**Description**

 - Fix bug on start date

**Jira Ticket:** https://bahmni.atlassian.net/jira/software/c/projects/BAH/boards/25?modal=detail&selectedIssue=BAH-3049

### DIFF
--- a/ui/app/reports/views/reports.html
+++ b/ui/app/reports/views/reports.html
@@ -45,7 +45,7 @@
             <tr ng-repeat="report in ::reportsRequiringDateRange" show-if-privilege="{{:: report.requiredPrivilege}}">
                 <td>{{:: report.name |translate }}</td>
                 <td class="reports-start-date">
-                    <input date-converter class="form-field start-date" type="date" max="{{report.stopDate | date:'yyyy-MM-dd'}}"
+                    <input date-converter class="form-field start-date" type="date" max="{{report.stopDate + 1 | date:'yyyy-MM-dd'}}"
                            ng-model="report.startDate">
                 </td>
                 <td class="reports-stop-date">


ui/app/reports/views/reports.html
```diff
@@ -45,7 +45,7 @@ <h2 class="section-title">{{::'REPORTS_TITLE_KEY' | translate}}</h2>
             <tr ng-repeat="report in ::reportsRequiringDateRange" show-if-privilege="{{:: report.requiredPrivilege}}">
                 <td>{{:: report.name |translate }}</td>
                 <td class="reports-start-date">
-                    <input date-converter class="form-field start-date" type="date" max="{{report.stopDate | date:'yyyy-MM-dd'}}"
+                    <input date-converter class="form-field start-date" type="date" max="{{report.stopDate + 1 | date:'yyyy-MM-dd'}}"
```
gsluthra, line 48: I didn't understand. What does this code change do. Also, it says "stopDate" + 1 for "startDate"? Is this as expected?

gsluthra: Does the max provide some check to ensure startDate is always <= stopDate, in which case, the +1 was needed for that?

Samridhi-98: Yes, correct
It was failing because of this input validation check startDate should be < stopDate thats why for Today date range startDate always set with undefined due to which it got fail so added  +1 .

gsluthra (approved):

ui/app/reports/views/reports.html
gokultw: The data value is set as 'max="Tue Sep 19 2023 12:29:14 GMT+0530 (India Standard Time)1"' instead of '2023-09-19.' It just concatenated the '1' and didn't add the day.

The AS-IS behavior is that if '2023-09-19' is set for the start and end date, then the 'min' and 'max' are also set to '2023-09-19.' So, I don't think we need to add one day.

@Samridhi-98 @gsluthra @priyanshu1602

PR #716, opened 2023-10-01
BAH 3049 | Samridhi | Fix start date undefined value

Samridhi-98:


ui/app/reports/controllers/reportsController.js
```diff
@@ -76,6 +77,9 @@ angular.module('bahmni.reports')
                 if (!report.stopDate) {
                     msg.push("end date");
                 }
+                if ((report.startDate > report.stopDate)) {
+                    msg.push("start date can not be greater than stop date");
```
gsluthra, line 81: Seems to be hardcoded english message. Should come from translation.

ui/app/i18n/reports/locale_en.json
```diff
@@ -44,5 +44,6 @@
     "CHOOSE_ANSWER_FROM_DROPDOWN_LABEL": "Choose Answer",
     "DELETE_LABEL": "Delete",
     "EDIT_LABEL": "Edit",
-    "REPORT_DATE_RANGE": "Select Date Range"
+    "REPORT_DATE_RANGE": "Select Date Range",
+    "START_DATE_CANNOT_GREATER_THAN_STOP_DATE": "start date can not be greater than stop date"
```
gsluthra, line 48: Can you rephrase this to "Later" instead of "Greater"?

_Start date cannot be later than the End Date_.

ui/test/unit/reports/controllers/reportsController.spec.js
```diff
@@ -108,6 +108,82 @@ describe("ReportsController", function () {
...
+            dateRangeType: new Date(new Date().setDate(currentDate.getDate() - 6))
+        };
+
+        scope.setDefault('dateRangeType', 'reportsRequiringDateRange');
+
+        expect(_.keys(rootScope.default.reportsRequiringDateRange).length).toBe(3);
+        expect(rootScope.reportsRequiringDateRange[0].startDate.getDate()).toBe(rootScope.default.reportsRequiringDateRange.startDate.getDate());
+        expect(rootScope.reportsRequiringDateRange[0].startDate.getMonth()).toBe(currentDate.getMonth());
+        expect(rootScope.reportsRequiringDateRange[0].stopDate.getDate()).toBe(currentDate.getDate());
+        expect(rootScope.reportsRequiringDateRange[0].stopDate.getMonth()).toBe(currentDate.getMonth());
```
gsluthra, line 171: If the currentDate is `2-Oct-2023 (Mon)`, then for Menu option `Last 7 days`, the dates selected automatically should be: `26-Sep-2023 (Tue)` to `2-Oct-2023 (Mon)`. In this case the startDate month = Sep (and not Oct). But I think your test is checking startDate.getMonth() to be currentDate.getMonth(). That isn't right I think. This test will break on certain days I think. Please reconfirm.

It would have been best if somehow tests could be written independent of actual date (today), and instead could be passed something like: If Current Date is X, then option Last 7 Days should show DateA and DateB, and so on. I am guessing that will require a little refactoring if your code, to take currentDate as a parameter, and then do calculations.

gsluthra (changes requested):
Please see my comment on Test cases.

ui/test/unit/reports/controllers/reportsController.spec.js
gsluthra: Maybe you can otherwise try and use Jasmine clock to mock the current Date. See this ChatGPT discussion I just did regarding this code block (scroll to end for Jasmine clock): https://chat.openai.com/share/42d88843-c43c-4dfc-81bc-6ef80438efbf

Samridhi-98:
> Please see my comment on Test cases.

sure, I'll check.

PR #749, opened 2023-10-18
Sowmya | fix alerts showing for unsaved observation even after all observations have been saved

sowmya-AS:


ui/app/common/concept-set/directives/formControls.js
```diff
@@ -92,11 +92,7 @@ angular.module('bahmni.common.conceptSet')
                         if (formObservation.value && formObservation.value.uuid && consultationObservation.value && consultationObservation.value.uuid) {
                             return (consultationObservation.value.uuid === formObservation.value.uuid) ? false : true;
                         } else {
-                            if (angular.isNumber(formObservation.value)) {
-                                formObservation.value = formObservation.value.toString();
-                            }
-
-                            return (consultationObservation.value === formObservation.value) ? false : true;
+                            return (consultationObservation.value === formObservation.value || (formObservation.value && consultationObservation.value === formObservation.value.toString())) ? false : true;
```
Phanindra-tw, line 95: Can we do something like this 
`return !((consultationObservation.value === formObservation.value || (formObservation.value && consultationObservation.value === formObservation.value.toString())))`
instead of ternary operator handling it?

Phanindra-tw (approved):

PR #788, opened 2023-11-29
BAH-3175 | Added Previous Month option in Bahmni Reports pre-filter

deeptirawat1510:


deeptirawat1510:
@gsluthra we have added test cases covering all the scenarios mentioned in the jira ticket, can you pls review this?

ui/test/unit/reports/controllers/reportsController.spec.js
```diff
@@ -94,18 +102,42 @@ describe("ReportsController", function () {
...
         scope.setDefault('dateRangeType', 'reportsRequiringDateRange');
-        scope.setDefault('startDate', 'reportsRequiringDateRange');
-        scope.setDefault('stopDate', 'reportsRequiringDateRange');
 
         expect(_.keys(rootScope.default.reportsRequiringDateRange).length).toBe(3);
         expect((rootScope.default.reportsRequiringDateRange.dateRangeType).getDate()).toBe(new Date().getDate());
-        expect(rootScope.reportsRequiringDateRange[0].startDate).toBe(rootScope.default.reportsRequiringDateRange.startDate);
-        expect(rootScope.reportsRequiringDateRange[0].stopDate).toBe(rootScope.default.reportsRequiringDateRange.stopDate);
+        expect(rootScope.reportsRequiringDateRange[0].startDate.getDate()).toBe(new Date().getDate());
+        expect(rootScope.reportsRequiringDateRange[0].stopDate.getDate()).toBe(new Date().getDate());
```
gsluthra, line 112: I didn't understand this assertion. The startdate and stop data is the same?

ui/test/unit/reports/controllers/reportsController.spec.js
```diff
@@ -94,18 +102,42 @@ describe("ReportsController", function () {
...
+    });
+
+    const previousMonthTestCases = [
```
gsluthra, line 115: Since all these scenarios have been added into a single datastructure, when this test fails, how does one know which use-case failed? Is there some log which prints the expected/calculated date?

ui/test/unit/reports/controllers/reportsController.spec.js
```diff
@@ -94,18 +102,42 @@ describe("ReportsController", function () {
...
+        { currentDate: '1-Dec-2022', expectedStartDate: '1-Nov-2022', expectedStopDate: '30-Nov-2022' },
```
gsluthra, line 116: Thanks for covering all the corner cases! appreciate this.

ui/test/unit/reports/controllers/reportsController.spec.js
deeptirawat1510: @gsluthra , when test case fails it would print the specific test case name for example- " should return previous date and month when ${currentDate}`" . Here current date would be the specific date.

gsluthra: No, maybe this is sufficient.

ui/test/unit/reports/controllers/reportsController.spec.js
gsluthra: The test was for `today`, so makes sense. thanks!

gsluthra (approved):

PR #792, opened 2023-12-01
BAH-3181 | Configuration support for concept name label to be shown on FSN if specified

angshu:
Configuration support in Obs display control ("displayNameType"), so that the concept name labels can be shown in FSN. If nothing specified default behavior of showing in SN (or FSN in SN is not available) is retained.
For example: 

 `
"basicVitals": {
        "translationKey": "XYZ",
        "type": "observation",
        "isObservation": true,
        "displayOrder": 6,
        "dashboardConfig": {
          "conceptNames": [
            "Height (cm)",
            "Weight (kg)",
            "Body mass index",
            "BMI Status"
          ],
          "showDetailsButton": true,
          "numberOfVisits": 3,
          "displayNameType": "FSN"
        },
        "hideEmptyDisplayControl": true
      }
`

This also resolves the bug where in non-implementation locale usage by user (e,g. default locale = en, user locale = french), the obs display control would not show non-matching concept names, even though the API returned the right data.

### DIFF
--- a/ui/app/common/displaycontrols/observation/directives/bahmniObservation.js
+++ b/ui/app/common/displaycontrols/observation/directives/bahmniObservation.js
@@ -19,7 +19,8 @@ angular.module('bahmni.common.displaycontrol.observation')
                     if ($scope.config.conceptNames) {
                         observations = _.filter(observations, function (observation) {
                             return _.some($scope.config.conceptNames, function (conceptName) {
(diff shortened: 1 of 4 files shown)


mohan-13:
@angshu In reference to this [PR](https://github.com/Bahmni/bahmni-core/pull/236) in bahmni-core, we would need to pass on the current user selected locale while making `/bahmnicore/observations` API call from frontend right ? Or is that handled in some other way ?

PR #810, opened 2024-01-17
BAH-3439 || Read also the timezone of birthtime.

harshkumar8789:
BAH-3439 || Change the timezone of birthtime.

### DIFF
--- a/ui/app/common/util/dateUtil.js
+++ b/ui/app/common/util/dateUtil.js
@@ -270,7 +270,7 @@ Bahmni.Common.Util.DateUtil = {
     },
 
     parseLongDateToServerFormat: function (longDate) {
-        return longDate ? moment(longDate).format("YYYY-MM-DDTHH:mm:ss.SSS") : null;
+        return longDate ? moment(longDate).format("YYYY-MM-DDTHH:mm:ss.SSSZZ") : null;
     },
 
     parseServerDateToDate: function (longDate) {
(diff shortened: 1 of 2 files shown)


gsluthra:
A unit test related to Operation Theatre Surgical blocks is failing. Plz check @harshkumar8789 @rahu1ramesh

ui/app/common/util/dateUtil.js
```diff
@@ -270,7 +270,7 @@ Bahmni.Common.Util.DateUtil = {
     },
 
     parseLongDateToServerFormat: function (longDate) {
```
gsluthra, line 272: This seems to be a method in dateUtils.js (which is a common utility), and hence there will possibly be multiple users of this method. Have you verified that all callers of this method will continue to work fine with this change. Alternatively introduce a new method called, parseLongDateToServerFormatWithTimezone( ) so, only callers which care about this information call it.

harshkumar8789: @gsluthra we did manual check with @gokultw . it seems that this changes doesnot effect, anywhere this method has been used.

gokultw: we identified one issue in appointments but that is an existing issue.

gsluthra (approved):

ui/app/common/util/dateUtil.js
rahu1ramesh: @gokultw do we have a card created for the appointments issue?

gokultw: I need to check with CURE team.

gsluthra:
@akhilmalhotra : The PR name is confusing. It is basically saying, don't ignore the timezone information while reading birthtime, so that on UI the same time is shown later also, as was entered while registering the patient.

PR #819, opened 2024-01-24
BAH-3502 | Refactor. Add a configuration property to hide provider name

deeptirawat1510:
This PR adds a config property to hide provider name in print pdf. The property `hideProviderName` will be added to printing section of each tab in visit.json file.

Example:
openmrs/apps/clinical/visit.json
```
"printing": {
            "title": "Bahmni",
            "header": "Discharge Summary",
            "logo": "../images/bahmniLogo.png",
            "hideProviderName": true
        },
```

### DIFF
--- a/ui/app/clinical/common/views/visitTabPrint.html
+++ b/ui/app/clinical/common/views/visitTabPrint.html
@@ -19,6 +19,6 @@
         </div>
         <div ng-include="currentVisitUrl"></div>
     </div>
-    <div class="provider-info">{{provider[0].name}}</div>
+    <div ng-if="visitTabConfig.getPrintConfigForCurrentTab().hideProviderNane" class="provider-info">{{provider[0].name}}</div>
 </body>
 </html>


ui/app/clinical/common/views/visitTabPrint.html
```diff
@@ -19,6 +19,6 @@ <h1>
         </div>
         <div ng-include="currentVisitUrl"></div>
     </div>
-    <div class="provider-info">{{provider[0].name}}</div>
+    <div ng-if="visitTabConfig.getPrintConfigForCurrentTab().hideProviderNane" class="provider-info">{{provider[0].name}}</div>
```
gsluthra, line 22: variable name has a typo: `hideProviderName`.

gsluthra: Also, if this variable is not present, then it will be "false" in which case by default the provider name will get hidden. Would prefer that by default this is shown (if variable is not defined). Can u invert the logic?

PR #843, opened 2024-02-13
BAH-3512 | Add. Sort Lab Orders By Name and Result Columns in Descending Order By Date

parvathy00:
JIRA -> [BAH-3512](https://bahmni.atlassian.net/browse/BAH-3512)

In this PR, I have added the logic to sort lab orders by test name(alphabetically) and to sort lab result columns in descending order by date(i.e, latestFirst), based on the value of configuration property.

Screenshot 2024-02-13 at 9 44 44 AM

### DIFF
--- a/ui/app/clinical/displaycontrols/investigationresults/directives/investigationResults.js
+++ b/ui/app/clinical/displaycontrols/investigationresults/directives/investigationResults.js
@@ -6,7 +6,8 @@ angular.module('bahmni.clinical')
             var defaultParams = {
                 showTable: true,
                 showChart: true,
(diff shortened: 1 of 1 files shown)


gsluthra:
If the new fields are not defined (using older config), will the code still work, with defaults?

parvathy00:
> If the new fields are not defined (using older config), will the code still work, with defaults?

Yes @gsluthra, I have added "chartConfig: {}" as the default param. So, it will work fine with default if new fields are not defined in the config file.

gsluthra (approved):

PR #847, opened 2024-02-15
BAH-3567 | Fix. Unexpected Server Error When Adding an Existing Diagnosis 

parvathy00:
JIRA -> [BAH-3567](https://bahmni.atlassian.net/browse/BAH-3567)

In this PR, we had fixed unexpected server error message when adding an existing Diagnosis by showing proper error message for duplicate diagnosis.
Screenshot 2024-02-15 at 2 51 06 PM

### DIFF
--- a/ui/app/clinical/consultation/controllers/diagnosisController.js
+++ b/ui/app/clinical/consultation/controllers/diagnosisController.js
@@ -265,22 +265,22 @@ angular.module('bahmni.clinical')
             $scope.checkInvalidDiagnoses = function () {
                 $scope.errorMessage = "";
                 $scope.consultation.newlyAddedDiagnoses.forEach(function (diagnosis) {
-                    if (isInvalidDiagnosis(diagnosis) || isDuplicateDiagnosis(diagnosis)) {
+                    if (isDuplicateWithNewlyAddedDiagnosis(diagnosis) || isDuplicateWithSavedDiagnosis(diagnosis)) {
                         $scope.errorMessage = "{{'CLINICAL_DUPLICATE_DIAGNOSIS_ERROR_MESSAGE' | translate }}";
                     }
                 });
             };
 
-            var isInvalidDiagnosis = function (diagnosis) {
+            var isDuplicateWithNewlyAddedDiagnosis = function (diagnosis) {
                 var codedAnswers = _.map(_.remove(_.map($scope.consultation.newlyAddedDiagnoses, 'codedAnswer'), undefined), function (answer) {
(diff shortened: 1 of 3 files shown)


ui/app/clinical/consultation/controllers/diagnosisController.js
```diff
@@ -265,7 +265,7 @@ angular.module('bahmni.clinical')
             $scope.checkInvalidDiagnoses = function () {
                 $scope.errorMessage = "";
                 $scope.consultation.newlyAddedDiagnoses.forEach(function (diagnosis) {
-                    if (isInvalidDiagnosis(diagnosis)) {
+                    if (isInvalidDiagnosis(diagnosis) || isDuplicateDiagnosis(diagnosis)) {
```
gsluthra, line 268: The error message to be shown in case of inValidDiagnosis() and duplicateDiagnosis() is the same?

ui/app/clinical/consultation/controllers/diagnosisController.js
```diff
@@ -280,9 +280,20 @@ angular.module('bahmni.clinical')
                 return diagnosis.invalid;
             };
 
+            var isDuplicateDiagnosis = function (diagnosis) {
```
gsluthra, line 283: Can you please add a unit test for this method?

ui/app/clinical/consultation/controllers/diagnosisController.js
parvathy00: Yes @gsluthra. In both cases, we are checking for duplicate entries. isInvalidDiagnosis() checks for duplicates in newlyAddedDiagnoses whereas isDuplicateDiagnosis() compares newlyAddedDiagnoses with savedDiagnosesFromCurrentEncounter.

gsluthra: Then should we merge the two methods into one, so that here, it only reads it as "isDuplicateDiagnosis()"? Or rename the isInvalidDiagnosis() to: isDuplicateWithNewlyAddedDiagnosis(), and the second method to isDuplicateWithSavedDiagnosis()?

parvathy00: @gsluthra We have renamed the methods, since merging these two functions are causing some inconsistencies.

gsluthra (approved):

gsluthra:
Once merged and deployed on QA env, please re-test it once before moving to Ready-for-QA lane. thanks!

PR #856, opened 2024-02-22
BAH-3522 | Add. Date Time Parser In Local Time

rahu1ramesh:
PR -> [BAH-3522](https://bahmni.atlassian.net/browse/BAH-3522)

This PR introduces functions to convert a dateTime from UTC to local time in the front-end dateTime format. Additionally, it adds a constant to specify formats globally.

### DIFF
--- a/ui/app/common/constants.js
+++ b/ui/app/common/constants.js
@@ -69,6 +69,8 @@ Bahmni.Common = Bahmni.Common || {};
         dateFormat: "dd/mm/yyyy",
         dateDisplayFormat: "DD-MMM-YYYY",
         timeDisplayFormat: "hh:mm",
+        clientTimeDisplayFormat: "hh:mm A",
+        clientDateDisplayFormat: "DD MMM YYYY",
         emrapiDiagnosisUrl: EMRAPI + "/diagnosis",
         bahmniDiagnosisUrl: BAHMNI_CORE + "/diagnosis/search",
         emrapiDiagnosisLimit: Bahmni.Common.Constants && Bahmni.Common.Constants.emrapiDiagnosisLimit || 20,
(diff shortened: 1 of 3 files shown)


ui/test/unit/ot/controller/moveSurgicalAppointmentController.spec.js
```diff
@@ -156,7 +156,7 @@ describe("moveSurgicalAppointmentController", function () {
         expect(scope.availableBlocks.length).toBe(1);
         expect(scope.availableBlocks[0].uuid).toBe("cdcf3c4b-6149-4a69-8113-97f651fae025");
         expect(scope.availableSurgicalBlocksForGivenDate.length).toBe(1);
-        expect(scope.availableSurgicalBlocksForGivenDate[0].displayName).toBe("Hanna Janho1, OT 2 (5:30 am - 8:30 am)");
+        expect(scope.availableSurgicalBlocksForGivenDate[0].displayName).toBe("Hanna Janho1, OT 2 (05:30 AM - 08:30 AM)");
```
gsluthra, line 159: Is the format on UI changing from `5:30pm` to `05:30pm`? Ideally the 0 prefix is only used in 24hour format, not in 12-hour format. We should try and keep the time on UI to `5:30pm`.

gsluthra: https://momentjs.com/docs/#/displaying/

I think we need to use format as: `h:mm a` (to get time to show as: `5:30 pm`) instead of using format `hh:mm A`.

rahu1ramesh: Will make the necessary update, is there any reason why `h:mm a` is preferred over `hh:mm A`?

gsluthra (approved):

PR #863, opened 2024-03-02
BAH-3644 | Add. Workflow To Pull Translations

rahu1ramesh:
This workflow is designed to retrieve translations from Transifex once every month in case resources are updated.

### DIFF
--- a/.github/workflows/pull_translations.yml
+++ b/.github/workflows/pull_translations.yml
@@ -1,0 +1,43 @@
+name: Pull Translations from Transifex
+
+on:
+  schedule:
+    - cron: '0 0 1 * *' 
+  workflow_dispatch:
+
+jobs:
+  pull-translations:
+    name: Pull Translations from Transifex
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Pull Translations from Transifex
+        run: |
+          bash .github/scripts/transifex.sh pull
+        env:
+          TX_TOKEN: ${{ secrets.TX_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_translations.yml
```diff
@@ -0,0 +1,43 @@
...
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Pull Translations from Transifex
+        run: |
+          bash .github/scripts/transifex.sh pull
```
deepthi-mantena, line 21: @rahu1ramesh may be consider enabling automatic exit on error here and also a error message if pull operation failed.Since this is an external service like Transifex, pull might fail due to various reasons its better we add error handling .

rahu1ramesh: @deepthi-mantena makes sense. I can introduce a variable named `success` that the script will update. In case of a failed Transifex operation, I'll set `success` to false and conditionally trigger an `exit 1;`. This will automatically halt the execution of the rest of the workflow.

rahu1ramesh: @deepthi-mantena I tried the above approach, but it doesn't seem to offer significant improvements over the existing code behavior. Currently, the [code](https://github.com/Bahmni/openmrs-module-bahmniapps/blob/cc29534f7b8ad41a7fd54e0fa1207a65bba82bfe/.github/scripts/transifex.sh#L32) handles exceptions in a way that terminates the workflow with an exit statement upon failure. I also verified this behavior by testing it on one of my repositories.

Screenshot 2024-03-05 at 19 37 51

deepthi-mantena (approved):

PR #903, opened 2024-04-11
Bindu, Kavitha,Phani | BAH-3092 | Show primary diagnoses in OT list view 

binduak:
JIRA Link -> [https://bahmni.atlassian.net/browse/BAH-3092](https://bahmni.atlassian.net/browse/BAH-3092)
Config PR link -> [https://github.com/Bahmni/default-config/pull/132](https://github.com/Bahmni/default-config/pull/132)


rahu1ramesh (changes requested):
In both the OT list and print views should display only the primary diagnosis, provided the global property is not null.

ui/app/ot/views/listView.html
rahu1ramesh, line 1: The primary diagnosis column remains visible when `obs.conceptMappingsForOT` is set, regardless of whether `additionalCustomParam` is configured. This makes it challenging for the implementer to hide the column from view.

kavitha-sundararajan: added check for the global property to show the column

rahu1ramesh (approved):

PR #913, opened 2024-04-23
BAH-3545 | Add. Favicon To Home Page

rahu1ramesh:
JIRA -> [BAH-3545](https://bahmni.atlassian.net/browse/BAH-3545)

### DIFF
--- a/package/docker/index.html
+++ b/package/docker/index.html
@@ -5,6 +5,7 @@
 	<meta charset="UTF-8" />
 	<meta name="robots" content="noindex,nofollow" />
 	<meta name="googlebot" content="noindex,nofollow" />
+	<link rel="icon" type="image/x-icon" href="/bahmni/favicon.ico">
 	<title>Bahmni</title>
 	<style>
 		html {


umair-fayaz (changes requested):
I am noticing that we have a bit of redundancy with this `index.html` file being present in both the proxy service and the bahmni-web service. I believe we can streamline this process and avoid maintaining the same file in multiple places.

Instead of duplicating the `index.html` file, I suggest we centralize it within the bahmni-web service. Since the bahmni-web service is already serving this file, we can remove it from the proxy container. 

To ensure that requests to the root path "/" are directed to bahmni-web, we can update the proxy configuration to include a ProxyPass rule for "/" pointing to bahmni-web. 

Let me know if you have any thoughts or suggestions on this approach.

mohan-13:
> I am noticing that we have a bit of redundancy with this `index.html` file being present in both the proxy service and the bahmni-web service. I believe we can streamline this process and avoid maintaining the same file in multiple places.
> 
> Instead of duplicating the `index.html` file, I suggest we centralize it within the bahmni-web service. Since the bahmni-web service is already serving this file, we can remove it from the proxy container.
> 
> To ensure that requests to the root path "/" are directed to bahmni-web, we can update the proxy configuration to include a ProxyPass rule for "/" pointing to bahmni-web.
> 
> Let me know if you have any thoughts or suggestions on this approach.

@umair-fayaz This index.html from bahmni-web is used only by k8s environment where proxy image is not used and in docker compose based deployments proxy container serves and thats the reason for the redundancy

umair-fayaz:
> > I am noticing that we have a bit of redundancy with this `index.html` file being present in both the proxy service and the bahmni-web service. I believe we can streamline this process and avoid maintaining the same file in multiple places.
> > Instead of duplicating the `index.html` file, I suggest we centralize it within the bahmni-web service. Since the bahmni-web service is already serving this file, we can remove it from the proxy container.
> > To ensure that requests to the root path "/" are directed to bahmni-web, we can update the proxy configuration to include a ProxyPass rule for "/" pointing to bahmni-web.
> > Let me know if you have any thoughts or suggestions on this approach.
> 
> @umair-fayaz This index.html from bahmni-web is used only by k8s environment where proxy image is not used and in docker compose based deployments proxy container serves and thats the reason for the redundancy

Yes, @mohan-13 . While I understand that the index.html from bahmni-web is specifically used in Kubernetes environments where the proxy image is not utilized, my suggestion was aimed at streamlining the process even for docker-compose based deployments by loading the index.html from bahmni-web.
Eliminating redundancy in this manner could simplify maintenance. Considering that logos and icons for this index.html are fetched from bahmni-web, the proxy already relies on bahmni-web for these assets. So, integrating the index.html loading from bahmni-web in docker-compose setups could enhance consistency and reduce duplication. Do you foresee any challenges with this approach?

mohan-13:
> > > I am noticing that we have a bit of redundancy with this `index.html` file being present in both the proxy service and the bahmni-web service. I believe we can streamline this process and avoid maintaining the same file in multiple places.
> > > Instead of duplicating the `index.html` file, I suggest we centralize it within the bahmni-web service. Since the bahmni-web service is already serving this file, we can remove it from the proxy container.
> > > To ensure that requests to the root path "/" are directed to bahmni-web, we can update the proxy configuration to include a ProxyPass rule for "/" pointing to bahmni-web.
> > > Let me know if you have any thoughts or suggestions on this approach.
> > 
> > 
> > @umair-fayaz This index.html from bahmni-web is used only by k8s environment where proxy image is not used and in docker compose based deployments proxy container serves and thats the reason for the redundancy
> 
> Yes, @mohan-13 . While I understand that the index.html from bahmni-web is specifically used in Kubernetes environments where the proxy image is not utilized, my suggestion was aimed at streamlining the process even for docker-compose based deployments by loading the index.html from bahmni-web. Eliminating redundancy in this manner could simplify maintenance. Considering that logos and icons for this index.html are fetched from bahmni-web, the proxy already relies on bahmni-web for these assets. So, integrating the index.html loading from bahmni-web in docker-compose setups could enhance consistency and reduce duplication. Do you foresee any challenges with this approach?

I agree. But not sure how all the redirections would behave when the root `/` is proxied to bahmni-web. We need to validate all the redirection scenarios before making the change

PR #916, opened 2024-04-24
add. trivyignore for html-minifier

abinaya-u:


umair-fayaz (changes requested):
Can we add a little more description in the form of comment explaining why this was ignored

abinaya-u:
> Can we add a little more description in the form of comment explaining why this was ignored

please check now

umair-fayaz (approved):

PR #946, opened 2024-05-17
BAH-3844 | Enable IPD Feature in patient Dashboard

Arjun-Go:
JIRA - https://bahmni.atlassian.net/browse/BAH-3844

### DIFF
--- a/ui/app/adt/app.js
+++ b/ui/app/adt/app.js
@@ -51,11 +51,23 @@ angular.module('adt').config(['$stateProvider', '$httpProvider', '$urlRouterProv
             },
             abstract: true,
             views: {
+                'header': {
+                    templateUrl: 'views/headerAdt.html',
+                    controller: function ($scope, appService) {
+                        $scope.showClinicalDashboardLink = true;
+                        $scope.enableIPDFeature = appService.getAppDescriptor().getConfigValue('enableIPDFeature');
+                    }
+                },
                 'content': {
                     template: '<ui-view/>'
+                },
+                'additional-header': {
+                    templateUrl: '../common/patient/header/views/header.html',
+                    controller: function ($scope, appService) {
+                        $scope.enableIPDFeature = appService.getAppDescriptor().getConfigValue('enableIPDFeature');
+                    }
                 }
             },
-
             resolve: {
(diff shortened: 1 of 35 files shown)


ui/app/adt/app.js
```diff
@@ -53,18 +53,21 @@ angular.module('adt').config(['$stateProvider', '$httpProvider', '$urlRouterProv
...
                     }
                 },
                 'content': {
                     template: '<ui-view/>'
                 },
                 'additional-header': {
-                    templateUrl: '../common/patient/header/views/header.html'
+                    templateUrl: '../common/patient/header/views/header.html',
+                    controller: function ($scope, appService) {
+                        $scope.enableIPDFeature = appService.getAppDescriptor().getConfigValue('enableIPDFeature');
```
rahu1ramesh, line 67: Line 58 and 67 both perform the same task of initializing the scope variable. I am guessing we can do away with this, if we do a minor refactoring.

ui/app/clinical/common/controllers/visitController.js
```diff
@@ -11,12 +11,39 @@ angular.module('bahmni.clinical')
                 $state.current.views['print-content'].templateUrl;
             var showProviderInfo = appService.getAppDescriptor().getConfigValue('showProviderInfoinVisits');
             $scope.showProviderInfo = showProviderInfo !== false ? true : showProviderInfo;
+            var showPatientInfo = appService.getAppDescriptor().getConfigValue('showPatientInfoInVisits');
+            $scope.showPatientInformation = showPatientInfo !== false ? true : showPatientInfo;
```
rahu1ramesh, line 15: @Arjun-Go @kavitha-sundararajan The ternary operator used here is redundant. If showPatientInfo is false, it sets `$scope.showPatientInformation` to true, which might not be the intended behavior. Please validate the approach once and maybe try - `$scope.showPatientInformation = showPatientInfo !== false;` which simplifies the initialization further. Still if showPatientInfo is `undefined, null, or true, $scope.showPatientInformation` will be true. Do confirm if this is the intended behaviour.

binduak:
closing the PR as its no longer required

PR #947, opened 2024-05-19
BAH-3824 | Add. Transifex Resource For Micro-Frontend

rahu1ramesh:
JIRA -> [BAH-3824](https://bahmni.atlassian.net/browse/BAH-3824)

In this PR, a new resource for the micro-frontend subfolder has been created and added to Transifex. The tx-config file has been updated to include the newly added resource, ensuring that it is included in the localization process.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -60,6 +60,15 @@ resource_name          = IPD App
 replace_edited_strings = false
 keep_translations      = false
 
+[o:openmrs:p:bahmni:r:micro-frontend-app]
+file_filter            = micro-frontends/public/i18n/locale_<lang>.json
+source_file            = micro-frontends/public/i18n/locale_en.json
+type                   = KEYVALUEJSON
+minimum_perc           = 0
+resource_name          = Micro Frontend App
+replace_edited_strings = false
+keep_translations      = false
+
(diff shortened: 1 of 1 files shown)


.tx/config
```diff
@@ -60,6 +60,15 @@ resource_name          = IPD App
 replace_edited_strings = false
 keep_translations      = false
 
+[o:openmrs:p:bahmni:r:micro-frontend-app]
+file_filter            = micro-frontends/public/i18n/locale_<lang>.json
+source_file            = micro-frontends/public/i18n/locale_en.json
```
deepthi-mantena, line 65: If the source language change or if there could be multiple source files,should we consider making the source file dynamic or configurable as in above line(64)
@rahu1ramesh

rahu1ramesh: @deepthi-mantena To ensure better support from Transifex, it is recommended to keep the source language as English (though non-English source languages are supported). Throughout Bahmni, we have used English as the source language to keep the strings easily maintainable for developers. Transifex expects a single source file, so there is no need to make the source file definition dynamic.

deepthi-mantena (approved):

PR #988, opened 2024-07-19
BAH-4037 | Add. Configuration to support specifying orderTypeClassMap

mohan-13:
This PR enhances the Order Controller to display concepts containing a specific class of concepts for each order type. This is to support using the same grouping concept across order types. The configuration can be added in implementation-config `clinical/app.json` under config section. 
Example:
```json
"orderTypeClassMap" :{
            "Laboratory": ["Test",,"LabTest","LabSet"],
            "Radiology Orders": ["Radiology","Radiology/Imaging Procedure"],
            "Procedure Orders": ["Procedure"]
        },
```
**Backward Compatible: If the config is not specified, then all the concepts under the set is displayed.**

_Without Config:_
image

_With Config:_
image

### DIFF
--- a/ui/app/clinical/consultation/controllers/orderController.js
+++ b/ui/app/clinical/consultation/controllers/orderController.js
@@ -150,7 +150,9 @@ angular.module('bahmni.clinical')
                 $scope.activeTab.leftCategory = leftCategory;
                 $scope.activeTab.leftCategory.klass = "active";
 
-                $scope.activeTab.leftCategory.groups = $scope.getConceptClassesInSet(leftCategory);
+                var conceptClasses = $scope.getConceptClassesInSet(leftCategory);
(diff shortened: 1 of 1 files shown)


ui/app/clinical/consultation/controllers/orderController.js
```diff
@@ -94,7 +94,10 @@ angular.module('bahmni.clinical')
                 $scope.tabs = [];
                 _.forEach($scope.allOrdersTemplates, function (item) {
                     var conceptName = $scope.getName(item);
-                    $scope.tabs.push({name: conceptName ? conceptName : item.name.name, topLevelConcept: item.name.name});
+                    var tabName = conceptName ? conceptName : item.name.name;
```
angshu, line 97: can be just
> tabName = conceptName || item.name.name

angshu (approved):

PR #990, opened 2024-07-29
BAH-3952 | Refactor bed management. use location uuid instead of parentLocation uuid

SanoferSameera:


SanoferSameera:
the changes aren't needed as it is mandatory to follow two level location hierarchy in creating the beds

PR #1023, opened 2024-11-26
BAH-4118 | Fix. Extension Iframe height to be responsive

mohan-13:
This PR fixes the responsiveness issue with the extension iframe that is shown from the Patient Registration for linking external identifiers.
JIRA: https://bahmni.atlassian.net/browse/BAH-4118

Before:
image

After:
image


devsingh05298:
The code looks fine to proceed.

PR #1037, opened 2025-11-24
Categorized reports and did role based access

rodney-m:


angshu:
@rodney-m   Hi, is there a corresponding card for this on our board where we can find some details? You can create a card easily here https://bahmni.atlassian.net/jira/dashboards/10402 and provide the card number as part of the description in format - BAH-#CardNumber | brief description. 

I also notice that you are trying to perform certain checks - which seem to be not generic and maybe more specific for your implementations. - antenatal, epidemic, screening etc. Not sure we understand the applicability

PR #1042, opened 2026-02-12
BAH-4442 | Filtering out retired program attribute types.

angshu:


ravinderkabli (approved):
What's good

  - The implementation is minimal and correct. angular.forEach with an early return acts as continue, so retired entries are cleanly skipped.
  - Adding retired to the custom representation string is the right approach — it avoids over-fetching.
  - All existing tests were correctly updated to reflect the new query string.

Approving the PR

PR #1053, opened 2026-05-25
BAH-4708 | Apply Alpine package upgrades in Dockerfile

Anutw:
## Summary
- Add `RUN apk --no-cache upgrade` to `package/docker/Dockerfile` so the `httpd:2.4-alpine` base image picks up the latest Alpine security patches at build time.

## Test plan
- [ ] `docker build -f package/docker/Dockerfile .` completes successfully
- [ ] Resulting image starts and serves on port 8091 as before
- [ ] Verify package patches are applied (e.g. `apk info -v` inside the built image vs. base)

### DIFF
--- a/package/docker/Dockerfile
+++ b/package/docker/Dockerfile
@@ -1,5 +1,7 @@
 FROM httpd:2.4-alpine
 
+RUN apk --no-cache upgrade
+
 EXPOSE 8091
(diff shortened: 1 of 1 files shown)


mohan-13 (commented):
This is not needed, a rebuild of the image might solve, give httpd-2.4-alpine would have fixed all of these dependencies

sumaztwcode:
These changes are not required

PR #1057, opened 2026-05-29
BAH-4742 The deeplink for amendment note.

AvniTanwar26:
Deeplink Amend notes.

### DIFF
--- a/ui/app/clinical/common/controllers/visitController.js
+++ b/ui/app/clinical/common/controllers/visitController.js
@@ -41,6 +41,10 @@ angular.module('bahmni.clinical')
             if ($scope.visitSummary.stopDateTime === null) {
                 $scope.isIpdReadMode = false;
             }
+            var privileges = [];
+            if ($rootScope.currentUser) {
+                privileges = $rootScope.currentUser.privileges || [];
+            }
             $scope.ipdDashboard = {
                 hostData: {
                     patient: {uuid: $scope.patientUuid},
@@ -50,7 +54,8 @@ angular.module('bahmni.clinical')
                     currentUser: $rootScope.currentUser,
                     visitUuid: $scope.visitUuid,
                     isReadMode: $scope.isIpdReadMode,
-                    source: $location.search().source
+                    source: $location.search().source,
+                    privileges: privileges
                 },
                 hostApi: {
                     navigation: {
--- a/ui/app/clinical/common/controllers/visitController.js
+++ b/ui/app/clinical/common/controllers/visitController.js
@@ -41,10 +41,8 @@ angular.module('bahmni.clinical')
             if ($scope.visitSummary.stopDateTime === null) {
                 $scope.isIpdReadMode = false;
             }
-            var privileges = [];
-            if ($rootScope.currentUser) {
-                privileges = $rootScope.currentUser.privileges || [];
-            }
+            var privileges = ($rootScope.currentUser && $rootScope.currentUser.privileges) || [];
+
             $scope.ipdDashboard = {
                 hostData: {
                     patient: {uuid: $scope.patientUuid},
--- a/ui/app/common/patient-search/controllers/patientsListController.js
+++ b/ui/app/common/patient-search/controllers/patientsListController.js
@@ -232,6 +232,8 @@ angular.module('bahmni.common.patientSearch')
             $.extend(options, {
                 patientUuid: patient.uuid,
                 visitUuid: patient.activeVisitUuid || null,
+                noteUuid: patient.noteUuid || null,
+                medicationAdministrationEpoch: patient.medicationAdministrationEpoch || null,
                 encounterUuid: $stateParams.encounterUuid || 'active',
                 programUuid: patient.programUuid || null,
                 enrollment: patient.enrollment || null,
--- a/ui/app/common/patient-search/controllers/patientsListController.js
+++ b/ui/app/common/patient-search/controllers/patientsListController.js
@@ -232,7 +232,7 @@ angular.module('bahmni.common.patientSearch')
             $.extend(options, {
                 patientUuid: patient.uuid,
                 visitUuid: patient.activeVisitUuid || null,
-                noteUuid: patient.noteUuid || null,
+                medicationAdministrationUuid: patient.medicationAdministrationUuid || null,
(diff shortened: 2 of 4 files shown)


ravinderkabli (commented):
Review for BAH-4742

ui/app/clinical/common/controllers/visitController.js
```diff
@@ -50,7 +54,8 @@ angular.module('bahmni.clinical')
                     currentUser: $rootScope.currentUser,
                     visitUuid: $scope.visitUuid,
                     isReadMode: $scope.isIpdReadMode,
-                    source: $location.search().source
+                    source: $location.search().source,
```
ravinderkabli, line 55: blocking: The `mfe-ipd-dashboard` receives its input only through `host-data="ipdDashboard.hostData"` (see `visitIpd.html`), and `hostData` forwards only `source` here. The deeplink also carries `openAcknowledge`, `medicationAdministrationUuid`, and `medicationAdministrationEpoch`, but those are not forwarded — so the dashboard has no way to open the specific slot / show the acknowledge action (parent story BAH-4619 AC1).

Mirror the `source` pattern:
```js
source: $location.search().source,
openAcknowledge: $location.search().openAcknowledge || null,
medicationAdministrationUuid: $location.search().medicationAdministrationUuid || null,
medicationAdministrationEpoch: $location.search().medicationAdministrationEpoch || null,
privileges: privileges
```
If the MFE reads these from `window.location` directly, please confirm and this can be closed.

ui/app/clinical/common/controllers/visitController.js
```diff
@@ -50,7 +54,8 @@ angular.module('bahmni.clinical')
...
+                    privileges: privileges
```
ravinderkabli, line 56: suggestion: `hostData` already passes `currentUser` (line above), which contains `.privileges` — so this separate `privileges` field is redundant and can be read as `hostData.currentUser.privileges`. Sibling controllers (`adtController`, `careViewController`) pass `currentUser` whole without a standalone `privileges` field, so this also deviates from the established pattern.

Prefer dropping the extra field, or add a short comment if the MFE genuinely needs a flat top-level `privileges` array.

ui/app/clinical/common/controllers/visitController.js
```diff
@@ -41,6 +41,10 @@ angular.module('bahmni.clinical')
             if ($scope.visitSummary.stopDateTime === null) {
                 $scope.isIpdReadMode = false;
             }
+            var privileges = [];
+            if ($rootScope.currentUser) {
+                privileges = $rootScope.currentUser.privileges || [];
```
ravinderkabli: question: Minor — this 4-line guard could be a one-liner: `var privileges = ($rootScope.currentUser && $rootScope.currentUser.privileges) || [];`. Purely optional.

ui/app/common/patient-search/controllers/patientsListController.js
```diff
@@ -232,6 +232,8 @@ angular.module('bahmni.common.patientSearch')
             $.extend(options, {
                 patientUuid: patient.uuid,
                 visitUuid: patient.activeVisitUuid || null,
+                medicationAdministrationUuid: patient.medicationAdministrationUuid || null,
```
ravinderkabli, line 235: nit: These two new forward-URL params have no test coverage. Since they feed a deeplink template, consider a `forwardPatient` spec covering a patient with and without `medicationAdministrationUuid`/`Epoch`, asserting the options passed to `formatUrl(...)`.

ui/app/common/patient-search/controllers/patientsListController.js
```diff
@@ -232,6 +232,8 @@ angular.module('bahmni.common.patientSearch')
...
+                medicationAdministrationEpoch: patient.medicationAdministrationEpoch || null,
```
ravinderkabli, line 236: question: `patient.medicationAdministrationEpoch || null` coerces a `0` epoch (1970-01-01) to `null`. Effectively impossible clinically and consistent with the surrounding `|| null` convention, so probably fine — just flagging.

ui/test/unit/clinical/common/controllers/visitController.spec.js
```diff
@@ -413,4 +413,66 @@ describe('VisitController', function () {
         });
     });
 
+    function createController(rootScope) {
```
ravinderkabli: nit: `createController` re-instantiates `VisitController` on the same `scope` that `beforeEach` already built, so the controller is constructed twice per test (duplicate `$window` keydown / `$destroy` listeners), and it omits the `visitService` local that `beforeEach` provides. The privilege assertions themselves are correct.

Consider having `beforeEach` call `createController(rootScope)` to consolidate to a single instantiation path, and include `visitService` in the helper. (Also: the closing brace of the function is over-indented.)

ui/app/clinical/common/controllers/visitController.js
SooryaKumaranC-tw: Okay will update

ui/app/clinical/common/controllers/visitController.js
SooryaKumaranC-tw: Yes correct MFE reads through `window.location` only.

ui/app/clinical/common/controllers/visitController.js
SooryaKumaranC-tw: No code change needed. Verified the MFE consumes privileges as its own top-level context value in multiple places (CalendarRow.jsx, DrugChartSlider.jsx, DrugChartView.jsx — gates the "acknowledge medication note" action), independent of currentUser. Not redundant.

ui/app/common/patient-search/controllers/patientsListController.js
SooryaKumaranC-tw: Added 2 specs in patientsListController.spec.js covering both present and absent cases.

ui/test/unit/clinical/common/controllers/visitController.spec.js
SooryaKumaranC-tw: Sure will remove this

ravinderkabli (approved):

PR #1058, opened 2026-06-01
BAH-4755 [Feature UI] Bahmni-V2: New Home for All, Privilege-Gated Old Registration Tile and Clinical Queue Tabs

lingeswaranTW:


ravinderkabli (approved):
Looks fine

PR #1063, opened 2026-06-12
BAH-4755 Home page as config

lingeswaranTW:


ui/app/home/loginInitialization.js
```diff
@@ -33,7 +33,23 @@ angular.module('bahmni.home')
...
+                loadConfigService.loadConfig(Bahmni.Common.Constants.baseUrl + "home/app.json").then(
+                    function (response) {
+                        var config = response.data && response.data.config;
+                        $rootScope.homeURL = (config && config.homeUrl) || Bahmni.Common.Constants.homeUrl;
+                        localStorage.setItem('homeUrl', $rootScope.homeURL);
+                        configDeferred.resolve();
+                    },
+                    function () {
+                        $rootScope.homeURL = Bahmni.Common.Constants.homeUrl;
+                        configDeferred.resolve();
```
ravinderkabli, line 47: suggestion: The error handler sets `$rootScope.homeURL` but doesn't call `localStorage.setItem()`. If a previous successful login cached a custom URL in `localStorage`, a transient config-load failure here leaves that stale value in place — other modules reading `localStorage` in their `.run()` blocks will use the old URL while this module uses the fallback constant.

Add `localStorage.setItem('homeUrl', $rootScope.homeURL);` here to keep `localStorage` in sync.

ravinderkabli (changes requested):
## Review for BAH-4755

ui/app/home/loginInitialization.js
arshiyaTW2021: Thanks for calling this out. the error handler in loginInitialization.js was setting $rootScope.homeURL to the fallback constant but not calling localStorage.setItem, leaving a potentially stale URL cached from a previous login. Added localStorage.setItem('homeUrl', $rootScope.homeURL) to the error handler so it stays consistent with the success path.

ravinderkabli: ok

ravinderkabli (approved):
Looks fine

PR #1079, opened 2026-07-22
BAH-4847 | Fix bed management home button to navigate to configured homeURL

ROHIT-V-TW:
Changed homeBackLink to fetch homeURL from localStorage to be consistent with other apps on Bahmni Clinical


ravinderkabli (approved):
Looks fine

PR #1080, opened 2026-07-28
BAH-4830 OT Enhancement

lingeswaranTW:


ui/app/ot/views/calendarSurgicalAppointment.html
```diff
@@ -8,7 +8,7 @@
         <p ng-if="attributes.estTimeHours || attributes.estTimeMinutes"> {{ 'OT_SURGICAL_APPOINTMENT_ESTIMATED_TIME'| translate}} - {{attributes.estTimeHours || 0 }} hrs(s): {{attributes.estTimeMinutes || 0}} min(s) </p>
         <p ng-if="surgicalAppointment.primaryDiagnosis"> {{ 'OT_SURGICAL_APPOINTMENT_PRIMARY_DIAGNOSIS' | translate}}  - {{surgicalAppointment.primaryDiagnosis}} </p>
         <p ng-if="attributes.procedure"> {{ 'OT_SURGICAL_APPOINTMENT_PROCEDURE' | translate}} - {{attributes.procedure}} </p>
-        <p ng-if="attributes.anaesthetist"> {{ 'OT_SURGICAL_APPOINTMENT_ANAESTHETIST'| translate}} - {{attributes.anaesthetist}} </p>
+        <p ng-if="attributes.anesthesic"> {{ 'OT_SURGICAL_APPOINTMENT_ANAESTHETIST'| translate}} - {{attributes.anesthesic}} </p>
```
ravinderkabli: blocking: `attributes.anesthesic` is always undefined — the attribute type is named `anaesthetist` everywhere else (helper `getSurgicalAttributes`/`getDefaultAttributeTranslations`, the list-view spec, and the untouched `surgicalAppointmentDialog.html` all use `anaesthetist`). So the anaesthetist line will never render on the calendar. Revert to `attributes.anaesthetist`, or rename consistently across the attribute name, helper map, all templates, locale key, and tests.

ui/app/ot/views/surgicalAppointment.html
```diff
@@ -28,44 +28,56 @@ <h4 class="appointment-title" ng-show="ngDialogData.sortWeight >= 0 ">{{'EDIT_SU
...
                 <p ng-switch-when="patientWeight">
-                    <label class="surgeon-block surgeon-text">{{'OT_SURGICAL_APPOINTMENT_PATIENT_WEIGHT' | translate}}</label>
-                    <input type="text" id="{{attribute.surgicalAppointmentAttributeType.patientWeight}}" ng-model="attribute.value">
+                    <label class="surgeon-block surgeon-text">{{'OT_SURGICAL_APPOINTMENT_PATIENT_WEIGHT' | translate}}<span class="asterick" ng-if="requiredConceptAttributes.indexOf(attributeTypeName) !== -1">*</span></label>
+                    <input type="text" id="{{attribute.surgicalAppointmentAttributeType.patientWeight}}" ng-model="attribute.value" ng-required="requiredConceptAttributes.indexOf(attributeTypeName) !== -1">
                 </p>
-                <p ng-switch-when="anaesthetist">
-                    <label class="surgeon-block surgeon-text">{{'OT_SURGICAL_APPOINTMENT_ANAESTHETIST' | translate}}</label>
-                    <input type="text" id="{{attribute.surgicalAppointmentAttributeType.anaesthetist}}" ng-model="attribute.value">
+                <p ng-switch-when="anesthesic">
```
ravinderkabli: blocking: Same incomplete rename as the calendar view — `ng-switch-when="anesthesic"` no longer matches the attribute (type name is `anaesthetist`), so the field falls through to the generic `ng-switch-default` text input instead of its dedicated field. Revert to `anaesthetist` or rename everywhere consistently.

ui/app/ot/index.html
```diff
@@ -147,13 +147,16 @@
...
     <script src="../common/concept-set/services/conceptGroupFormatService.js"></script>
+    <script src="../common/concept-set/services/conceptService.js"></script>
     <script src="../common/concept-set/mappers/observationMapper.js"></script>
     <script src="../common/concept-set/models/conceptSetObservation.js"></script>
     <script src="../common/concept-set/models/booleanObservation.js"></script>
     <script src="../common/concept-set/models/observationNode.js"></script>
     <script src="../common/concept-set/models/tabularObservations.js"></script>
     <script src="../common/concept-set/models/multiSelectObservations.js"></script>
     <script src="../common/concept-set/models/customRepresentationBuilder.js"></script>
+    <script src="../clinical/consultation/services/observationFormService.js"></script>
```
ravinderkabli: suggestion: This file does not exist (`ui/app/clinical/consultation/services/observationFormService.js` is absent — only a stale spec remains), so this resolves to a 404. It's also unreferenced anywhere in the OT module; the concept lookups use `conceptService` (added above). Please remove this line (and the stray extra blank line).

ui/app/ot/controller/surgicalBlockController.js
```diff
@@ -17,7 +17,9 @@ angular.module('bahmni.ot')
                     surgicalAppointments: []
                 };
                 $scope.configuredSurgeryAttributeNames = appService.getAppDescriptor().getConfigValue("surgeryAttributes");
-                $scope.defaultAttributeTranslations = surgicalAppointmentHelper.getDefaultAttributeTranslations();
+                var listViewTranslations = appService.getAppDescriptor().getConfigValue("listViewTranslations") || [];
+                $scope.defaultAttributeTranslations = surgicalAppointmentHelper.getDefaultAttributeTranslations(listViewTranslations);
+                $scope.conceptFormatAttributeName = surgicalAppointmentHelper.getConceptFormatAttributeName ? surgicalAppointmentHelper.getConceptFormatAttributeName() : Bahmni.OT.Constants.conceptFormatAttributeName;
```
ravinderkabli: suggestion: `surgicalAppointmentHelper.getConceptFormatAttributeName` doesn't exist — that method lives on the new `otUtils` factory. The ternary is therefore always false and this always resolves to the hardcoded constant `'Blood Transfusion Requested for Surgery?'`, defeating the dynamic format-based discovery used in the other three consumers. Inject `otUtils` and call `otUtils.getConceptFormatAttributeName()`, then drop `conceptFormatAttributeName` from constants.js.

ui/app/ot/mappers/SurgicalBlockMapper.js
```diff
@@ -38,9 +38,57 @@ Bahmni.OT.SurgicalBlockMapper = function () {
...
+        var latestValue = null;
+        var latestDate = null;
+        _.each(observations, function (obs) {
+            if (!obs || !obs.display) {
+                return;
+            }
+            if (obs.concept && obs.concept.display === conceptDisplayName) {
+                var currentObsDate = obs.obsDatetime;
+                var parsedDate = currentObsDate && Bahmni.Common.Util.DateUtil.parseServerDateToDate(currentObsDate);
+                if (!latestDate || (parsedDate && parsedDate > latestDate)) {
```
ravinderkabli: suggestion: When a newer obs of the same concept has a truthy `display` but a falsy `obs.value`, `latestDate` advances while `latestValue` keeps the older value — so the returned `{value, date}` come from different observations, and the validity check then runs against the newer date (can keep a stale value alive). Update `latestValue` and `latestDate` atomically from the same obs.

ui/app/i18n/ot/locale_en.json
```diff
@@ -66,6 +66,10 @@
     "OT_ENTER_YOUR_NOTES": "enter your notes",
     "OT_ADD_KEY": "Add",
     "OT_CANCEL_KEY": "Cancel",
+    "OT_SELECT_OPTION_KEY": "Select an option",
+    "OT_YES_KEY": "Yes",
+    "OT_NO_KEY": "No",
+    "REQUIRED_FIELD_MESSAGE": "Please make a selection",
```
ravinderkabli: suggestion (AC5): `REQUIRED_FIELD_MESSAGE` is added here but never referenced in any template. Required fields only get an asterisk + native `ng-required`; the user is never shown the prompt message the AC calls for. Wire it into the form (e.g. `<span ng-show="form[name].$invalid && form.$submitted">{{'REQUIRED_FIELD_MESSAGE' | translate}}</span>`) or remove the key if asterisk-only is intended.

ui/app/i18n/ot/locale_fr.json
```diff
@@ -0,0 +1,14 @@
+{
+    "OT_SELECT_OPTION_KEY": "Sélectionner une option",
+    "Requested by Surgeon": "Demandé par le chirurgien",
```
ravinderkabli: suggestion (AC6): The concept attribute label (`Blood Transfusion Requested for Surgery?`) and these coded-answer values are translated only in French. There are no matching keys in `locale_en.json` and no `translationKey` mapping for the concept label, so English/other locales rely on angular-translate's raw-key fallback. Add the label + answer-value keys to `locale_en.json` (and other supported locales).

ui/app/ot/constants.js
```diff
@@ -18,7 +18,10 @@ Bahmni.OT.Constants = (function () {
         weekDays: {"Monday": 1, "Tuesday": 2, "Wednesday": 3, "Thursday": 4, "Friday": 5, "Saturday": 6, "Sunday": 7 },
         defaultWeekStartDayName: 'Sunday',
         providerSurgicalAttributeFormat: 'org.openmrs.Provider',
-        notesUrl: RESTWS_V1 + '/notes'
+        notesUrl: RESTWS_V1 + '/notes',
+        conceptFormatAttributeName: 'Blood Transfusion Requested for Surgery?',
+        defaultObservationValidityDays: 30,
+        notApplicableValues: ['Not Applicable', 'Non applicable']
```
ravinderkabli: nit: `notApplicableValues` hardcodes localized display strings and is compared against the *translated* value in the views to blank the cell — this only works for these two languages and breaks for any other locale or wording change. Prefer matching on a stable concept UUID/canonical key, or drive the suppression set from config.

ui/app/ot/mappers/SurgicalBlockMapper.js
```diff
@@ -38,9 +38,57 @@ Bahmni.OT.SurgicalBlockMapper = function () {
...
+                    }
+                    if (parsedDate) {
+                        latestDate = parsedDate;
+                    }
+                }
+            }
+        });
+        if (latestDate && validityDays) {
+            var now = new Date();
+            var diffDays = (now - latestDate) / (1000 * 60 * 60 * 24);
```
ravinderkabli: nit: This day-diff math duplicates `Bahmni.Common.Util.DateUtil.diffInDays` / `diffInDaysRegardlessOfTime` (already in scope — `parseServerDateToDate` is used just above). Reusing it also normalizes the fractional-day behavior (this keeps fractions, no `Math.floor`).

ui/app/ot/mappers/SurgicalBlockMapper.js
```diff
@@ -38,9 +38,57 @@ Bahmni.OT.SurgicalBlockMapper = function () {
...
+        _.each(observations, function (obs) {
+            if (!obs || !obs.display) {
+                return;
+            }
+            if (obs.concept && obs.concept.display === conceptDisplayName) {
+                var currentObsDate = obs.obsDatetime;
+                var parsedDate = currentObsDate && Bahmni.Common.Util.DateUtil.parseServerDateToDate(currentObsDate);
+                if (!latestDate || (parsedDate && parsedDate > latestDate)) {
+                    if (obs.value) {
+                        latestValue = obs.value.display || obs.value;
```
ravinderkabli: nit: If `obs.value` is an object without a `display` property, this falls back to the whole object and would render as `[object Object]` in the list cell. Consider `obs.value.display || obs.value.name || ''` (or coerce).

ui/app/ot/controller/newSurgicalAppointmentController.js
```diff
@@ -10,18 +10,47 @@
...
+    .controller('NewSurgicalAppointmentController', ['$scope', '$rootScope', '$q', '$window', 'patientService', 'surgicalAppointmentService', 'messagingService', 'programService', 'appService', 'ngDialog', 'spinner', 'queryService', 'programHelper', 'surgicalAppointmentHelper', 'conceptService', 'otUtils',
+        function ($scope, $rootScope, $q, $window, patientService, surgicalAppointmentService, messagingService, programService, appService, ngDialog, spinner, queryService, programHelper, surgicalAppointmentHelper, conceptService, otUtils) {
             var init = function () {
                 $scope.configuredSurgeryAttributeNames = appService.getAppDescriptor().getConfigValue("surgeryAttributes");
                 $scope.selectedPatient = $scope.ngDialogData && $scope.ngDialogData.patient;
                 $scope.patient = $scope.ngDialogData && $scope.ngDialogData.patient && ($scope.ngDialogData.patient.value || $scope.ngDialogData.patient.display);
                 $scope.otherSurgeons = _.cloneDeep($scope.surgeons);
+                $scope.currentLanguage = $window.localStorage["NG_TRANSLATE_LANG_KEY"] || "en";
```
ravinderkabli: nit: `$scope.currentLanguage` is never read anywhere in the module (it only forced a `localStorage` stub in the test). Remove it, or use it if intended.

ui/app/ot/mappers/SurgicalBlockMapper.js
```diff
@@ -38,9 +38,57 @@ Bahmni.OT.SurgicalBlockMapper = function () {
...
+            }
+        }
+        return { value: latestValue, date: latestDate };
+    };
+
+    var mapObservations = function (patientObservations, conceptConfigs) {
+        var result = {};
+        _.each(conceptConfigs, function (config) {
+            result[config.conceptName] = mapConcepts(patientObservations, config.conceptName, config.validityDays)
+                || { date: null, value: "" };
```
ravinderkabli: nit: `mapConcepts` always returns an object, so this `|| { date: null, value: "" }` fallback is unreachable dead code.

ui/app/i18n/ot/locale_en.json
```diff
@@ -66,6 +66,10 @@
     "OT_ENTER_YOUR_NOTES": "enter your notes",
     "OT_ADD_KEY": "Add",
     "OT_CANCEL_KEY": "Cancel",
+    "OT_SELECT_OPTION_KEY": "Select an option",
+    "OT_YES_KEY": "Yes",
```
ravinderkabli: nit: `OT_YES_KEY` and `OT_NO_KEY` are added but not referenced anywhere in the PR or module. Remove them, or wire in if a Yes/No dropdown was intended. (`OT_SELECT_OPTION_KEY` above is used.)

ui/app/i18n/ot/locale_fr.json
```diff
@@ -0,0 +1,14 @@
+{
```
ravinderkabli, line 1: nit: This new French file covers only ~13 of the ~130 OT keys; the rest fall back to English. Additive and fine (no override risk), but please confirm partial French coverage is intended for this feature.

ui/app/ot/controller/listViewController.js
```diff
@@ -10,42 +10,106 @@
...
 angular.module('bahmni.ot')
-    .controller('listViewController', ['$scope', '$rootScope', '$q', 'spinner', 'surgicalAppointmentService', 'appService', 'surgicalAppointmentHelper', 'surgicalBlockFilter', 'printer',
-        function ($scope, $rootScope, $q, spinner, surgicalAppointmentService, appService, surgicalAppointmentHelper, surgicalBlockFilter, printer) {
+    .controller('listViewController', ['$scope', '$rootScope', '$q', '$window', 'spinner', 'surgicalAppointmentService', 'appService', 'surgicalAppointmentHelper', 'surgicalBlockFilter', 'printer', 'otUtils',
+        function ($scope, $rootScope, $q, $window, spinner, surgicalAppointmentService, appService, surgicalAppointmentHelper, surgicalBlockFilter, printer, otUtils) {
             var startDatetime = moment($scope.viewDate).toDate();
             var surgicalBlockMapper = new Bahmni.OT.SurgicalBlockMapper();
             var endDatetime = moment(startDatetime).endOf('day').toDate();
-            $scope.defaultAttributeTranslations = surgicalAppointmentHelper.getDefaultAttributeTranslations();
+            var listViewTranslations = appService.getAppDescriptor().getConfigValue("listViewTranslations") || [];
```
ravinderkabli: nit: Translation overrides are declared under two differently-named keys — `listViewTranslations` here vs `additionalSurgeryAttributeTranslations` in newSurgicalAppointmentController — plus `listViewObservationColumns[].translationKey`, all feeding one `defaultAttributeTranslations` Map. This is the duplication the JIRA flagged; consider consolidating to a single key/path.

ui/app/ot/initialization.js
```diff
@@ -31,4 +31,27 @@ angular.module('bahmni.ot').factory('initialization', ['$rootScope', '$q', 'surg
         };
         return spinner.forPromise(authenticator.authenticateUser().then(initApp).then(loadConfigPromise));
     }
-]);
+]).factory('otUtils', ['$rootScope', function ($rootScope) {
```
ravinderkabli, line 34: question: `otUtils` exposes both `getConceptFormatAttributeName` (first concept attr) and `getConceptFormatAttributeNames` (all). The form handles all concept attributes while list/calendar use only the first — if a site configures more than one `org.openmrs.Concept` attribute, only the first shows as a column/badge. Intended? Also, would this be better placed on `surgicalAppointmentHelper` (which already filters attributeTypes) than in the bootstrap file — that would also have avoided the wrong-helper reference in surgicalBlockController.

ui/app/ot/directives/otCalendarSurgicalAppointment.js
```diff
@@ -75,7 +79,8 @@ angular.module('bahmni.ot')
                 operationTheatre: "=",
                 heightPerMin: "=",
                 backgroundColor: "=",
-                filterParams: "="
+                filterParams: "=",
+                conceptFormatAttributeName: "=?"
```
ravinderkabli: question: This optional `=?` binding is never passed by the parent template (the link fn falls back to `otUtils`), so the scope entry is currently dead weight. Drop it and rely on the fallback, or bind it from the parent if calendar/list are meant to share a resolved value.

ui/app/ot/views/listView.html
```diff
@@ -18,22 +18,29 @@
...
-                {{appointment.derivedAttributes.expectedStartDate | date : 'EEEE'}}
+                {{appointment.derivedAttributes.patientName}}
             </td>
             <td>
-                {{appointment.derivedAttributes.expectedStartDate | bahmniDate}}
+                {{appointment.status}}
             </td>
             <td>
-                {{appointment.derivedAttributes.patientIdentifier}}
+                {{conceptFormatAttributeName && appointment.surgicalAppointmentAttributes[conceptFormatAttributeName] ? (conceptFormatDropdownConstants.indexOf((appointment.surgicalAppointmentAttributes[conceptFormatAttributeName].value | translate)) === -1 ? (appointment.surgicalAppointmentAttributes[conceptFormatAttributeName].value | translate) : '') : ''}}
```
ravinderkabli: question: This blanks the concept cell whenever the value matches `notApplicableValues`. AC4 says selected values should display correctly — is hiding "Not Applicable" intended product behavior? (See also the constants.js note about the hardcoded localized strings.)

ui/app/ot/index.html
PoojaSastry-TW: Removed this unused reference to observationFormService.js.

ui/app/ot/controller/surgicalBlockController.js
PoojaSastry-TW: Fixed by injecting otUtils (consistent with calendarViewController, listViewController, and newSurgicalAppointmentController) and calling otUtils.getConceptFormatAttributeName() directly. Also removed the now-unused conceptFormatAttributeName constant from constants.js.

ui/app/ot/mappers/SurgicalBlockMapper.js
PoojaSastry-TW: latestValue and latestDate are now updated atomically from the same obs.

ui/app/i18n/ot/locale_en.json
PoojaSastry-TW: Removed the unused REQUIRED_FIELD_MESSAGE key.

PoojaSastry-TW: Also removed the unused OT_YES_KEY and OT_NO_KEY.

ui/app/i18n/ot/locale_fr.json
PoojaSastry-TW: Removing unused Requested by Surgeon key and Not Applicable keys.

ui/app/ot/constants.js
PoojaSastry-TW: Removed notApplicableValues from constants.js and dropped the conceptFormatDropdownConstants assignments from all three consumers (calendarViewController, listViewController, otCalendarSurgicalAppointment), and simplified listView.html to display the value directly without suppression (which was earlier being enforced for Not applicable values as per CURE requirement).

ui/app/ot/mappers/SurgicalBlockMapper.js
PoojaSastry-TW: Updated to reuse the existing Bahmni functionality.

ui/app/ot/mappers/SurgicalBlockMapper.js
PoojaSastry-TW: Fixed. Added obs.value.name as a fallback before the raw value, and used angular.isObject to coerce any remaining object without display or name to an empty string, preventing [object Object] from rendering in the list cell. Added a test covering this case.

ui/app/ot/controller/newSurgicalAppointmentController.js
PoojaSastry-TW: Removed $scope.currentLanguage along with the localStorage stub in the test that was only there to support it.

ui/app/i18n/ot/locale_en.json
PoojaSastry-TW: The YES and NO keys have been removed.

ui/app/ot/mappers/SurgicalBlockMapper.js
PoojaSastry-TW: Removed the unreachable fallback.

ui/app/i18n/ot/locale_fr.json
PoojaSastry-TW: Partial French coverage is intentional and scoped to keys introduced by this feature. Also removed deployment-specific translationKey values (OT_ANAESTHESIA_REVIEW_DATE, OT_ANAESTHESIA_REVIEW, OT_PAEDIATRIC_REVIEW_DATE, OT_PAEDIATRIC_REVIEW) from locale_en.json — since listViewObservationColumns is fully config-driven, each deployment defines its own translationKey values and corresponding locale entries.

ui/app/ot/controller/listViewController.js
PoojaSastry-TW: Already consolidated — all three consumers (surgicalBlockController, listViewController, newSurgicalAppointmentController) now read from a single surgeryAttributeTranslations config key and pass it through surgicalAppointmentHelper.getDefaultAttributeTranslations.

ui/app/ot/initialization.js
PoojaSastry-TW: In the OT List View, all concept format attributes are visible, the first gets the dedicated column, additional ones fall through into filteredSurgicalAttributeTypes and render as generic attribute columns. 

As for the calendar badge, this is a known gap and can be tracked as a separate improvement. The calendar badge inclusion has not been added as a contribution in the JIRA. 

Placement on otUtils vs surgicalAppointmentHelper: Valid point, moving it to surgicalAppointmentHelper would be cleaner since it already handles attribute type filtering. However, surgicalAppointmentHelper is currently a plain service with no $rootScope injection, while otUtils needs $rootScope.attributeTypes which is populated during app initialization.     
Moving it would require injecting $rootScope into surgicalAppointmentHelper. This is a reasonable refactor but we'd prefer to keep it as a separate tracked improvement rather than expanding the scope of this PR, since otUtils is working correctly and all consumers are now using it consistently.

ui/app/ot/directives/otCalendarSurgicalAppointment.js
PoojaSastry-TW: concept-format-attribute-name is never passed by the parent template (calendarSurgicalBlock.html), so the =? binding was dead weight. Removed it along with the if (!$scope.conceptFormatAttributeName) guard, otUtils.getConceptFormatAttributeName() is now called unconditionally as the sole source.

ui/app/ot/views/listView.html
PoojaSastry-TW: Fixed — notApplicableValues and the suppression logic have been removed, as this was a CURE specific requirement.

ui/app/ot/views/calendarSurgicalAppointment.html
PoojaSastry-TW: The change was relevant to CURE as anaesthetist was changed to anesthesic in the DB for a different requirement. Reverting to use attributes.anaesthetist here.

ui/app/ot/views/surgicalAppointment.html
PoojaSastry-TW: Reverted to use attributes.anaesthetist.

ravinderkabli (approved):
Looks fine

PR #1085, opened 2026-08-06
BAH-4952:Add Hyperlink Support for Label Controls in Forms

Nandinidasari60:
## Summary

- Adds hyperlink rendering for Label form controls in patient views
- Implements backward compatibility tests for hyperlink configuration
- Updates form definitions to support hyperlinkUrl property
- Moves hyperlinkAllowedDomains configuration to flat config structure
- Adds translation tests for hyperlink-enabled labels
- Supports runtime hyperlink validation in form rendering

## Behaviour

| Scenario | Before | After |
|---|---|---|
| Label with hyperlinkUrl | Plain text | Clickable hyperlink |
| Form with hyperlink label | N/A | Renders link properly |
| Configuration change | N/A | Takes effect immediately |
| Old form definitions | Still work | Backward compatible |
| URL domain check | N/A | Validates against whitelist |

## Configuration

Update form definitions to include hyperlink:
```json
{
  "type": "label",
  "value": "Click here",
  "properties": {
    "hyperlinkUrl": "https://who.int"
  }
}
```

## Files Changed

- `ui/app/views/patient/forms/labelControl.html` - Hyperlink rendering
- `test/unit/common/domain/ConceptTest.groovy` - Backward compatibility tests
- `test/unit/common/domain/FormBuilderTest.groovy` - Form definition tests
- `src/main/resources/liquibase/changelog/form-definitions-hyperlink.xml` - Migration
- `test/integration/forms/HyperlinkFormTest.groovy` - Integration tests

## Test plan

- [ ] Form with hyperlink label renders correctly
- [ ] Hyperlink is clickable and navigates correctly
- [ ] Domain validation prevents unauthorized URLs
- [ ] Old forms without hyperlinks still work (backward compat)
- [ ] Form data persists with hyperlink property
- [ ] URL with patient UUID tokens resolves correctly
- [ ] Hyperlink displays in patient views
- [ ] No regression in label rendering without hyperlink
- [ ] Database migration completes successfully
- [ ] Translations work for hyperlinked labels

### DIFF
--- a/micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.jsx
+++ b/micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.jsx
@@ -80,8 +80,12 @@ const EditObservationForm = (props) => {
 
                 const [latestForms, allowedDomainsData] = await Promise.all([
                     getLatestPublishedForms(),
-                    axios.get(GLOBAL_PROPERTY_URL, { params: { property: 'bahmni.forms.hyperlink.allowedDomains' } })
-                        .then(res => res.data || '')
+                    axios.get(GLOBAL_PROPERTY_URL, {
+                            params: { property: 'bahmni.forms.hyperlink.allowedDomains' },
+                            withCredentials: true,
+                            headers: { Accept: 'text/plain' }
+                        })
+                        .then(res => String(res.data || ''))
                         .catch(() => '')
                 ]);
 
--- a/micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.jsx
+++ b/micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.jsx
@@ -9,6 +9,7 @@
 
 import React, { useEffect, useState } from 'react';
 import PropTypes from "prop-types";
+import axios from 'axios';
 import { getLocale } from "../i18n/utils";
 import { getFormByFormName, getFormDetail, getFormTranslations } from "./EditObservationFormUtils";
 import { findByEncounterUuid } from '../../utils/FormDisplayControl/FormView';
@@ -79,8 +80,8 @@ const EditObservationForm = (props) => {
 
                 const [latestForms, allowedDomainsData] = await Promise.all([
                     getLatestPublishedForms(),
-                    fetch(GLOBAL_PROPERTY_URL + '?property=bahmni.forms.hyperlink.allowedDomains', { credentials: 'include' })
-                        .then(res => res.ok ? res.text() : '')
+                    axios.get(GLOBAL_PROPERTY_URL, { params: { property: 'bahmni.forms.hyperlink.allowedDomains' } })
+                        .then(res => res.data || '')
                         .catch(() => '')
                 ]);
 
--- a/micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.jsx
+++ b/micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.jsx
@@ -17,6 +17,7 @@ import { Modal, Loading } from 'carbon-components-react';
 import { FormattedMessage } from "react-intl";
 import { I18nProvider } from '../i18n/I18nProvider';
 import ErrorNotification from '../ErrorNotification/ErrorNotification';
+import { GLOBAL_PROPERTY_URL } from '../../constants';
 
 import "./EditObservationForm.scss";
 
@@ -28,15 +29,15 @@ const EditObservationForm = (props) => {
         />
     );
 
-    const { 
-        formName, 
+    const {
+        formName,
         formNameTranslations,
-        closeEditObservationForm, 
+        closeEditObservationForm,
         isEditFormLoading,
-        setEditFormLoading, 
-        patient, 
-        formData, 
-        encounterUuid, 
+        setEditFormLoading,
+        patient,
+        formData,
+        encounterUuid,
         consultationMapper,
         handleEditSave,
         editErrorMessage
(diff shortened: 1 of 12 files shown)


ui/app/common/concept-set/directives/formControls.js
```diff
@@ -29,48 +29,64 @@ angular.module('bahmni.common.conceptSet')
...
+                        console.error('Failed to fetch hyperlink allowed domains:', error);
+                        loadForm();
+                    });
 
                 $scope.$watch('form.collapseInnerSections', function () {
                     var collapse = $scope.form.collapseInnerSections && $scope.form.collapseInnerSections.value;
                     if (loadedFormDetails[formUuid]) {
-                        $scope.form.component = renderWithControls(loadedFormDetails[formUuid], formObservations,
-                            formUuid, collapse, $scope.patient, validateForm, locale, loadedFormTranslations[formUuid]);
+                        renderForm(loadedFormDetails[formUuid], loadedFormTranslations[formUuid]);
```
ravinderkabli, line 89: blocking: Collapse-toggle regression. `renderForm` closes over the outer-scope `collapse` (computed once at controller init). This `$watch` recomputes a local `collapse` but `renderForm(...)` ignores it and re-renders with the stale load-time value, so toggling collapse/expand inner sections no longer takes effect.

Pass collapse in explicitly: `var renderForm = function (formDetails, formTranslations, collapse) { ... }` and have this call site pass the newly-computed local `collapse`. Worth a unit test asserting the collapse arg (index 3) passed to `renderWithControls`.

ui/app/common/concept-set/directives/formControls.js
```diff
@@ -10,8 +10,8 @@
 'use strict';
 
 angular.module('bahmni.common.conceptSet')
-    .directive('formControls', ['formService', 'spinner', '$timeout', '$translate',
-        function (formService, spinner, $timeout, $translate) {
+    .directive('formControls', ['formService', 'spinner', '$timeout', '$translate', '$state', 'messagingService', 'appService', '$http', 'configurationService',
```
ravinderkabli: suggestion: `$state`, `messagingService`, `appService`, and `$http` are injected here but never referenced in the directive body (only `configurationService` is used). Please drop them from both the annotation array and the factory signature, and remove the corresponding unused mocks in the spec.

ui/app/common/concept-set/directives/formControls.js
```diff
@@ -29,48 +29,64 @@ angular.module('bahmni.common.conceptSet')
...
+                };
+
+                configurationService.getConfigurations(['hyperlinkAllowedDomains'])
+                    .then(function (configurations) {
+                        var domainsData = configurations.hyperlinkAllowedDomains || '';
+                        allowedDomains = (domainsData || '').split(',').map(function (d) { return d.trim(); }).filter(function (d) { return d; });
+                        loadForm();
+                    })
+                    .catch(function (error) {
+                        console.error('Failed to fetch hyperlink allowed domains:', error);
```
ravinderkabli: nit: This `console.error` fires on every load where the config fetch fails, but that's a handled fallback (empty whitelist). The React path swallows the same failure silently — consider dropping this or downgrading to a debug log for consistency.

micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.jsx
```diff
@@ -75,28 +77,35 @@ const EditObservationForm = (props) => {
             if( formData.length > 0 && encounterUuid !== null) {
                 const encounterTransaction = await findByEncounterUuid(encounterUuid);
                 setEncounter(consultationMapper.map(encounterTransaction));
-                
-                const latestForms = await getLatestPublishedForms();
+
+                const [latestForms, allowedDomainsData] = await Promise.all([
+                    getLatestPublishedForms(),
+                    axios.get(GLOBAL_PROPERTY_URL, { params: { property: 'bahmni.forms.hyperlink.allowedDomains' } })
```
ravinderkabli: suggestion: This raw axios call diverges from the established util pattern (`AllergyControlUtils.getNoKnownAllergyUuid`), which sets `withCredentials: true` and `headers: { Accept: 'text/plain' }`.

- Without `withCredentials`, this can fail auth in some deployments — silently, since the `.catch(() => '')` disables the whitelist with no signal.
- Without `Accept: text/plain`, axios may auto-parse a JSON-ish value, so `res.data` isn't a string and line 94's `.split` throws a TypeError (`|| ''` only guards falsy values).

Mirror the allergy util (add the header + `withCredentials`), coerce with `String(res.data || '')`, and consider extracting a `getHyperlinkAllowedDomains()` util to keep data-fetching out of the component (the Angular side already routes through `configurationService`).

ui/app/common/domain/services/configurationService.js
```diff
@@ -100,6 +100,20 @@ angular.module('bahmni.common.domain')
...
+        configurationFunctions.hyperlinkAllowedDomains = function () {
+            return $http.get(Bahmni.Common.Constants.globalPropertyUrl, {
+                params: {
+                    property: 'bahmni.forms.hyperlink.allowedDomains'
+                },
+                withCredentials: true,
+                transformResponse: [function (data) {
+                    return data;
+                }]
+            }).catch(function () {
```
ravinderkabli, line 112: nit: This `.catch` resolves to a bare string `''`, but `getConfigurations` reads `response.data`, so on error `configurations.hyperlinkAllowedDomains` becomes `undefined` (tolerated downstream by `|| ''`). For consistency with the sibling config functions, return `{ data: '' }` here, or drop the in-service catch and let the error bubble to `$q.all`.

ui/test/unit/clinical/consultation/directives/formControls.spec.js
```diff
@@ -90,6 +134,84 @@ describe("Form Controls", function () {
         expect(renderHelper.renderWithControlsCalledTimes).toBe(2);
     });
 
+    it("should set dirtyForm flag when changes are saved", function () {
```
ravinderkabli: nit: This test asserts `$state.dirtyForm` is falsy, but nothing sets it and there is no `$event:changes-saved` listener in the directive — so it passes trivially and verifies nothing. Please remove it (and the unused `$state` mock fields), or implement the behaviour the title implies.

micro-frontends/src/next-ui/Components/EditObservationForm/EditObservationForm.scss
```diff
@@ -74,4 +74,18 @@
 .bx--modal-content {
     padding-top: 0;
 }
+
+a[data-bahmni-hyperlink="true"] {
+    color: #0066cc;
```
ravinderkabli: nit: Hardcoded link colors (`#0066cc` here, `#663399` for `:visited`). Prefer a shared SCSS color token if one exists. The `[data-bahmni-hyperlink]` hover rule is also duplicated in `_observations.scss` — consider centralizing.

ravinderkabli (commented):
# Review for BAH-4952

ravinderkabli (approved):
Looks fine

PR #1089, opened 2026-08-18
Avni| BAH-4970| Ability to add URLs to specific forms from nursing tasks

AvniTanwar26:
Jira ticket: https://bahmni.atlassian.net/browse/BAH-4970


ui/app/clinical/consultation/controllers/conceptSetPageController.js
```diff
@@ -66,10 +66,41 @@ angular.module('bahmni.clinical')
...
 
-                    if (templateToBeOpened) {
+                    if (templateToBeOpened && !$stateParams.formUuid) {
                         openTemplate(templateToBeOpened);
                     }
                 }
+
+                var formUuidParam = $stateParams.formUuid;
+
+                $timeout(setupDirtyTracking, formUuidParam ? 1000 : 0);
```
ravinderkabli: blocking: `setupDirtyTracking` is not defined anywhere in the repo (only the call sites here and on line 90). This throws `ReferenceError` and aborts the observations load on every navigation. Define/inject it (along with the draft/dirty-tracking logic the new specs expect) or remove the call.

Separately, the `1000` ms delay is a fragile magic number — prefer an event/`$watch` on a real "form loaded" signal.

ui/app/clinical/consultation/controllers/conceptSetPageController.js
```diff
@@ -66,10 +66,41 @@ angular.module('bahmni.clinical')
...
+                    });
+                    if (targetForm) {
+                        if (!_.some($scope.consultation.selectedObsTemplate, function (t) { return t === targetForm; })) {
+                            targetForm.isAdded = true;
+                            $scope.consultation.selectedObsTemplate.push(targetForm);
+                        }
+                        $timeout(function () {
+                            $rootScope.$broadcast('event:openFormByUuid', { form: targetForm });
+                        }, 0);
+                        var deletedFormIds = getRootDeletedFormIds();
```
ravinderkabli, line 93: blocking: `getRootDeletedFormIds` is also undefined — same `ReferenceError` crash. Define/inject it or remove.

ui/app/clinical/consultation/controllers/conceptSetPageController.js
```diff
@@ -66,10 +66,41 @@ angular.module('bahmni.clinical')
...
+                        if (!_.includes(deletedFormIds, formUuidParam)) {
```
ravinderkabli, line 94: blocking: duplicate add + double broadcast. The block just above already pushed `targetForm` and scheduled `event:openFormByUuid` unconditionally, so this guarded block pushes/broadcasts a *second* time and the `deletedFormIds` guard is effectively bypassed. Net effect: the form opens twice.

Collapse to one path: check `deletedFormIds` first, then add-if-absent, then broadcast once.

ui/app/clinical/consultation/controllers/conceptSetPageController.js
```diff
@@ -66,10 +66,41 @@ angular.module('bahmni.clinical')
...
+                            if (!_.find($scope.consultation.selectedObsTemplate, function (t) { return t === targetForm; })) {
+                                targetForm.isAdded = true;
+                                $scope.consultation.selectedObsTemplate.push(targetForm);
+                            }
+                            $timeout(function () {
+                                $rootScope.$broadcast('event:openFormByUuid', { form: targetForm });
+                            }, 0);
+                        }
+                    } else {
+                        messagingService.showMessage('error', 'Form not found. Please contact your administrator.');
```
ravinderkabli, line 104: nit: hardcoded English string. Use `$translate.instant(...)` with a locale key (following the existing `CLINICAL_FORM_*` pattern) and add it to all `locale_*.json` files. Note AC2 also asks for the wording "Patient/Form not found" and a graceful page rather than an auto-dismissing toast.

ui/app/clinical/common/controllers/visitController.js
```diff
@@ -26,8 +26,10 @@ angular.module('bahmni.clinical')
             $scope.documentsPromise = encounterService.getEncountersForEncounterType($scope.patient.uuid, encounterTypeUuid).then(function (response) {
                 return new Bahmni.Clinical.PatientFileObservationsMapper().map(response.data.results);
             });
-            $scope.currentVisitUrl = $state.current.views['dashboard-content'].templateUrl ||
-                $state.current.views['print-content'].templateUrl;
+            var dashboardView = $state.current.views?.['dashboard-content'];
+            var printView = $state.current.views?.['print-content'];
+            var view = dashboardView?.templateUrl ? dashboardView : printView;
+            $scope.currentVisitUrl = view?.templateUrl ?? null;
```
ravinderkabli: blocking: `?.` / `??` (ES2020). The `ui` build minifies with uglify-js 2.8.29 and has no Babel; uglify 2.x can't parse optional chaining and the minify step will fail. No other source file in `ui/app` uses these operators. Use `_.get(...)` / explicit `&&` guards instead.

ui/app/clinical/dashboard/controllers/patientDashboardController.js
```diff
@@ -103,8 +103,7 @@ angular.module('bahmni.clinical')
                         $scope.diseaseTemplates = diseaseTemplate;
                         $scope.sectionGroups = dashboardModel.getSections($scope.diseaseTemplates);
                     });
-                $scope.currentDashboardTemplateUrl = $state.current.views['dashboard-content'] ?
-                    $state.current.views['dashboard-content'].templateUrl : $state.current.views['dashboard-content'];
+                $scope.currentDashboardTemplateUrl = $state.current.views?.['dashboard-content']?.templateUrl ?? $scope.currentDashboardTemplateUrl;
```
ravinderkabli: blocking: same `?.`/`??` build break as visitController.

Also a behavior change: `?? $scope.currentDashboardTemplateUrl` keeps the *previous* URL when `templateUrl` is undefined, whereas the original assigned `undefined` (cleared it). This can leave a stale template showing. Revert to ES5 and preserve the original clear-on-undefined semantics, e.g. `_.get($state, "current.views['dashboard-content'].templateUrl")`.

ui/app/clinical/app.js
```diff
@@ -182,6 +183,9 @@ angular.module('consultation')
                     }
                 }
             })
+            .state('patient.dashboard.show.observations.form', {
```
ravinderkabli, line 186: suggestion: this child state declares only a `url` (no `views`/`controller`). In-app navigation with `notify:false` keeps the parent view rendered, but please verify a *cold* browser load / shared deeplink to `.../form/:formUuid` actually instantiates the parent view and reads `$stateParams.formUuid` — that cold-load path is the whole point of AC2. Worth an explicit test.

ui/app/common/concept-set/directives/conceptSetGroup.js
```diff
@@ -178,12 +178,32 @@ angular.module('bahmni.common.conceptSet')
                 $scope.leftPanelConceptSet.atLeastOneValueIsSet = selectedConceptSet.hasSomeValue();
                 $scope.leftPanelConceptSet.isAdded = true;
                 $scope.consultation.lastvisited = selectedConceptSet.id || selectedConceptSet.formUuid;
+                var currentFormUuid = $state.params.formUuid;
+                if (selectedConceptSet.formUuid) {
+                    if (currentFormUuid !== selectedConceptSet.formUuid) {
+                        $state.go('patient.dashboard.show.observations.form',
```
ravinderkabli, line 184: blocking: `conceptSetGroup` lives in `bahmni.common.conceptSet` (a shared module), but this calls `$state.go('patient.dashboard.show.observations...')` — state names defined only in the clinical app (`clinical/app.js`). That couples a shared module to one app's route table and will no-op/error for any other consumer. Move the URL-sync into `ConceptSetPageController` (clinical layer) and keep this directive state-agnostic (e.g. emit an event).

ravinderkabli (commented):
## Review for BAH-4970

ui/app/common/concept-set/directives/conceptSetGroup.js
```diff
@@ -178,12 +178,32 @@ angular.module('bahmni.common.conceptSet')
...
+                            { formUuid: selectedConceptSet.formUuid },
+                            { notify: false, location: 'replace' }
+                        );
+                    }
+                } else {
+                    if (currentFormUuid) {
+                        $state.go('patient.dashboard.show.observations',
```
ravinderkabli, line 191: suggestion: `$state.go('...observations', {}, ...)` passes empty params, which can reset inherited params (patientUuid, conceptSetGroupName) to their defaults and drop patient context from the URL. Pass `{ formUuid: null }` and let ui-router inherit the rest.

ui/app/common/concept-set/directives/conceptSetGroup.js
```diff
@@ -178,12 +178,32 @@ angular.module('bahmni.common.conceptSet')
...
+                        );
+                    }
+                }
                 if ($rootScope.showLeftpanelToggle) {
                     $rootScope.showLeftpanelToggle = false;
                 }
                 $(window).scrollTop(0);
             };
 
+            $scope.$on('event:openFormByUuid', function (event, data) {
```
ravinderkabli, line 203: suggestion: the `$scope.$on(...)` deregistration fn isn't captured and there's no `$destroy` cleanup. Since it's a `$rootScope` broadcast, every live `ConceptSetGroupController` responds. Capture the return value and call it on `$destroy`, and guard that `data.form` belongs to this instance's `allTemplates`.

lingeswaranTW (dismissed):

lingeswaranTW (approved):

